PR #13, opened 2020-02-28
📦 atlas-experiment-table 💬 V2.0.0

alfonsomunozpomer:
**I strongly suggest that you review this PR from your IDE as if it were a new component with fresh eyes, trying to track the changes from the previous experiment table is going to be confusing, more than anything**

It all started with a simple requirement: set a minimum width to the experiments table.

While going after the story to allow passing initial values to the table filters, I noticed a small shortcoming of the table: if you resized your browser to a narrow window, the table wasn’t responsive at all and the contents became all garbled. Since I didn’t want to start doing a responsive view, I though an easy fix would be setting a minimum width applying a style using React’s CSS-in-JS, right? So I decided to take a closer look at Evergreen’s general documentation and source code, and found out that to do that properly in Evergreen we should take advantage of the functionality of [`ui-box`](https://github.com/segmentio/ui-box). However that lead me to see that [our proposed solution to make the table headers bigger didn’t use any Evergreen primitive](https://github.com/ebi-gene-expression-group/atlas-experiment-table/pull/16). This made that positioning and sizes of elements inside the table were always a bit off. So I decided to follow that earlier PR but with Evergreen’s [Heading](https://evergreen.segment.com/components/typography), [SearchInput](https://evergreen.segment.com/components/search-input), [Icon](https://evergreen.segment.com/components/icons), etc.

After this a second realisation came: we were no longer using Evergreen’s table filtering but our own logic, so we could abstract all that functionality to state-encapsulating component: `TableManager`. I added some [memoization](https://reactjs.org/docs/hooks-faq.html#how-to-memoize-calculations) so that paginating would be faster. While doing that I noticed that we were treating dropdown filters and table headers as different things, while we could abstract them to a single `filters` state field, greatly simplifying the filtering/ordering logic.

Another improvement is that headers can be searchable **and** sortable, combining capabilities as necessary.

Links from table elements are generated with a callback on the table row, which is a lot more flexible and easier to understand.

I think that it’s worth comparing the previous version of the experiments table in these regards as well:
- Code organisation, how to break up stuff into smaller, easier-to-test components (e.g. table headers). It’s always better to make components smaller than necessary rather than larger than necessary. Err on the side of size reduction!
- Dry out logic in filtering
- [Lodash](https://lodash.com/) is full of goodness that we can use
- Pagination logic was hard to follow, with a couple of simple assumptions it’s been simplified a lot (even there was an unnecessary prop)

Try the two example pages to get a feel of all the functionality.


packages/atlas-experiment-table/src/TableContent.js
```diff
@@ -1,141 +1,125 @@
...
+    dataRows, tableHeaders,
+    // Filter handling
+    filters, tableHeaderCellOnChange,
+    // Ordering
+    sortColumnIndex, ascendingOrder, tableHeaderCellOnClick,
+    // Row selection and action on selected rows
+    rowSelectionColumn, selectedRows, host, selectOnChange
+  }) =>
+    <div className={`row expanded`}>
+      <div className={`small-12 columns`}>
```
lingyun1010: I think to keep the code as general as possible, we could pass the div `className` provided by EBI using props as what we did for some other components. And we may use the default props to make the input props simpler. 

Here are a bunch of `className` in `TableFooter` and `TablePreamble` too.

packages/atlas-experiment-table/src/TableCell.js
```diff
@@ -0,0 +1,63 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import URI from 'urijs'
+
+import { Table, Text } from 'evergreen-ui'
+
+const TableCell = ({ dataRow, dataKey, image, linkTo, host, width }) => {
+  // If the column is an image put the image in the cell (or an unknown icon if the type is undefined)
```
lingyun1010: Personally, if the `if-else` logic level goes too deep (>2), I will use `if-else` or `switch` to make the code easy to be followed. 

Otherwise, line 8 comment is better to be moved below `const cellItem = image ?` line.

alfonsomunozpomer: Something like this?

```javascript
  let cellItem = null
  if (image) {
    // If the column is an image put the image in the cell (or an unknown icon if the type is undefined)
    if (image[dataRow[dataKey]]) {
      cellItem = <img {...image[dataRow[dataKey]]}/>
    }
    else {
      cellItem = `❔`
    }
  } else if (Array.isArray(dataRow[dataKey])) {
    // If the contents of the cell is an array expand it to a list
    cellItem =
      <ul>
        {dataRow[dataKey].map(
          (element, index) =>
            <li key={`${index}`}>
              {element}
            </li>
        )}
      </ul>
  } else {
    // Any other type (i.e. string, number) put it as-is in the cell
    cellItem = dataRow[dataKey]
  }
```

packages/atlas-experiment-table/src/TableContent.js
alfonsomunozpomer: I didn’t think of this and I think there’s value in being able to change the `className` attribute via props. Especially the outer one, since we want to have the table in the Cambridge Cell Atlas website that has a regular `row` (not `expanded`) layout.

packages/atlas-experiment-table/src/TableCell.js
alfonsomunozpomer: As a general rule I tend to favour the ternary operator because it returns a value (it’s an expression) instead of `if-else` which is a control flow statement. The ternary operator is thus more declarative and you can assign it to a `const`, whereas with `if` you need to declare your variable with `let` and your code is more imperative.

In any case, I can understand that in this case regular conditionals can make the code more readable and I changed the logic to use `if-else`. Here we can’t use `switch` because the expression under evaluation is not the same at each conditional. Maybe if JavaScript had [pattern matching](https://en.wikipedia.org/wiki/Pattern_matching)... [One can dream, though!](https://github.com/bramstein/funcy)

alfonsomunozpomer:
Have a look at the proposed changes. I’ve fixed the placement of the tooltip by setting it to the `left` of the info icon and resolved the duplicated JSON files by having the files in the `html` directory be a symbolic link from the tests, so that the files in `__test__` are now fixtures and the authoritative source.

I commented on your observations above. Remember to resolve the conversations if you’re happy with the outcome.

PR #15, opened 2020-02-28
📦 atlas-react-fetch-loader 💬 Allow to rename payload keys

alfonsomunozpomer:
Because the next version of the experiments table uses `dataRows` (which is a more agnostic, general name) but the endpoint gives you `experiments`, this is something that will be very useful. I think it can also come in handy, as it allows us to decouple React components from the payload.


packages/atlas-react-fetch-loader/src/FetchLoader.js
```diff
@@ -66,8 +66,12 @@ const withFetchLoader = (WrappedComponent) => {
           throw new Error(`${url} => ${response.status}`)
         }
 
+        const data = await response.json()
+        Object.keys(this.props.renameDataKeys)
+          .forEach(key => delete Object.assign(data, {[this.props.renameDataKeys[key]]: data[key] })[key])
```
alfonsomunozpomer: This is a one liner that does the following:
```javascript
// Add a new key which points at the old key contents: data[newKey] === data[oldKey]
Object.assign(data, { [newKey]: data[oldKey] })
// Remove old key
delete data[oldKey]
```

haideriqbal: Cool! A clean and simple!

PR #19, opened 2020-03-05
📦 atlas-experiment-table 💬 Feature #171628203 – Add routing

alfonsomunozpomer:
I created a new routing component in `TableManagerRoute` that encapsulates all the URL-wiring functionality. It’s then used in `TableManagerSpa`, which is the exported component.

I took the chance of debouncing the filtering function when it’s run after keystrokes to make the whole table more pleasant to use.


packages/atlas-experiment-table/src/TableManager.js
```diff
@@ -132,25 +149,32 @@ export default class TableManager extends React.Component {
...
-        _.chain(this.state.filters)
-          .omit(dataKey)  // Returns new object, this.state.filters isn’t mutated, can _.cloneDeep if needed though
-          .assign(
-            // An empty string means the filter has been deleted from the table header or All has been selected in a
-            // dropdown, so we add no keys to the filters object
-            value.trim() === `` ? {} : { [dataKey]: value })
-          .value(),
-      currentPage: 1
-    })
+  updateFilters(dataKey, value, debounce = true) {
```
lingyun1010: If `debounce` is always true, do we still need this variable?

packages/atlas-experiment-table/webpack.config.js
```diff
@@ -57,8 +57,8 @@ module.exports = {
   devServer: {
     port: 9000,
     contentBase: path.resolve(__dirname, `html`),
-    publicPath: commonPublicPath
+    publicPath: commonPublicPath,
     // Add if developing a SPA to redirect non-matching routes known by WDS (i.e. no document in /html) to the router
-    // historyApiFallback: true
+    historyApiFallback: true
```
lingyun1010: Please also add a text exception for `TableManagerSpa.js`.

packages/atlas-experiment-table/src/TableManager.js
alfonsomunozpomer: It’s false when updating the dropdowns (have a look at how the `dropdownOnChange` prop is used in `TablePreamble`). In that case we don’t want to delaythe filtering.

packages/atlas-experiment-table/webpack.config.js
alfonsomunozpomer: What do you mean?

lingyun1010: Since we do not have test for TableManagerSpa.js , so we could add an entry in config to say that this file is out of tests.

PR #27, opened 2020-03-12
📦 atlas-experiment-table 💬 Bugfix #171747234 – Handle intial values in dropdowns

alfonsomunozpomer:
We were setting an initial value through `defaultValue` which isn’t part of the HTML spec, so I suspect that the dropdowns were on uncontrolled mode and were half-working at best.


packages/atlas-experiment-table/src/TableManager.js
```diff
@@ -95,20 +95,19 @@ export default class TableManager extends React.Component {
       currentPage: 1,
       rowsPerPage: this.rowsPerPageOptions[0],
       // Fields in the filters/headers *must not be* arrays or objects, so the spread creates good clones
-      dropdownFiltersOptions:
-        props.dropdownFilters.reduce((acc, dropdownFilter) => {
-          acc[dropdownFilter.dataKey] =
+      dropdownFilters:
+        props.dropdownFilters.map(dropdownFilter => ({
```
haideriqbal: 👍 easier to follow now!

packages/atlas-experiment-table/html/render.js
```diff
@@ -1,14 +1,14 @@
...
 }
 
-const renderSpa = (options, target) => {
-  ReactDOM.render(<TableManagerSpa { ...options } />, document.getElementById(target))
+const renderRouter = (options, target) => {
+  ReactDOM.render(<TableManagerRouter { ...options } />, document.getElementById(target))
 }
 
-export { render, renderSpa }
+export { render, renderRouter }
```
haideriqbal: 🆒 makes it easier to know now.

PR #31, opened 2020-03-13
📦 atlas-experiment-table 💬 Feature #170000065 – Add dialogue window for file types

lingyun1010:
Add a popup window with check boxes for file type selection before downloading.


packages/atlas-experiment-table/html/index-sc.html
```diff
@@ -88,7 +104,7 @@
             width: 0.5
           },
           dataRows: data,
-          host: 'https://www.ebi.ac.uk/gxa/sc/'
+          host: 'http://localhost:8080/gxa/sc/'
```
haideriqbal: I guess it is better to keep it `https://www.ebi.ac.uk/gxa/sc`. I understand that your backend changes might not have been merged yet but for future purposes, it would be better.

packages/atlas-experiment-table/__test__/TableContent.test.js
```diff
@@ -11,7 +11,7 @@ import SelectionTableHeaderCell from '../src/head/SelectionTableHeaderCell'
 
 import randomString from 'random-string'
 
-import { getRandomInt, bulkTableHeaders } from './TestUtils'
+import {getRandomInt, bulkTableHeaders, downloadFileTypes} from './TestUtils'
```
haideriqbal: Please add space near curly braces.

packages/atlas-experiment-table/html/index-sc.html
lingyun1010: Restored to `www.eb.ac.uk`.

lingyun1010:
Regarding the propTypes of `downloadFileTypes`, as it is not required for bulk experiment table for now, I removed the requirement rather than adding an empty array.

We used `react-popup.css` in web-single-cell for the feedback popup window, so I think we could use the same style sheet.

packages/atlas-experiment-table/src/TableManager.js
```diff
@@ -58,7 +58,7 @@ export default class TableManager extends React.Component {
         description: PropTypes.string,
         id: PropTypes.string
       })
-    ).isRequired,
+    ),
```
alfonsomunozpomer: As a general practice, if a prop isn’t required it should have a default value in `defaultProps`.

packages/atlas-experiment-table/src/head/SelectionTableHeaderCell.js
```diff
@@ -27,7 +27,7 @@ class SelectionTableHeaderCell extends React.Component {
     super(props)
 
     this.state = {
-      fileTypes: props.downloadFileTypes.map(fileType => fileType.id)
+      fileTypes: props.downloadFileTypes && props.downloadFileTypes.map(fileType => fileType.id)
```
alfonsomunozpomer: By giving a default prop value you can remove the conditional check altogether, and your logic remains clean and straightforward. The fewer edge cases, the better.

packages/atlas-experiment-table/src/head/SelectionTableHeaderCell.js
```diff
@@ -111,7 +111,7 @@ SelectionTableHeaderCell.propTypes = {
       description: PropTypes.string,
       id: PropTypes.string
     })
-  ).isRequired,
```
alfonsomunozpomer: Remember again to provide default values for non-required props.

alfonsomunozpomer:
Also, with a monorepo you should be careful with the scope of your branch. After `npx lerna bootstrap` the `package-lock.json` file in `scxa-gene-search-form` was updated and you included it in 016097896ba106ee7409804913b383525ee34e95. Your branch should be limited to `atlas-experiment-table`. Discard anything else to keep things clean and tidy.

alfonsomunozpomer:
@lingyun1010 Please have a look at the refactored code.

The biggest change is the code reorganisation and separation of concerns. In the current version the table has a prop, `rowSelectionColumn`, with a field `tableHeaderCellOnClick` that is meant to hold a function that initiates the download process. In that way, any logic that is wired in there doesn’t interfere with the table’s own code. It should be a completely separate module. I had taken a shortcut by defaulting to the function exported by `downloadHelper` because it was exactly the same for bulk and single cell, but since now there’s some different behaviour by virtue of the file type logic, it was the right time to use it.

The most interesting part, I believe, is the `ExperimentDownloadModule` that does all the rendering and prepartion of the tooltip. It requires an extra DOM node as a placeholder, but it allows for a complete split between the table and the download function.

I’ve reimplemented the `Prompt` component as a functional React component with hooks and simplified the checkbox logic with Lodash’s `xor` function. Use it every time you have a multi-selection checkbox control in React. It’s the same as the experiment row selection.

Lastly the `validateAndDownloadExperimentFiles` is exported as a curried version. It would be the same as using [Lodash’s curry](https://lodash.com/docs#curry):
```javascript
const _validateAndDownloadExperimentFiles = _.curry(validateAndDownloadExperimentFiles)(host, _, [])
```

However I prefer explicit closures, I think they’re easier to read.
```

How about if you think a few tests for `ExperimentDownloadModule`?

alfonsomunozpomer:
I added some tests for the download module. Since we do a render within the method to register the tooltip, we need to do everything through the DOM, which makes the tests a bit hard to read, very imperative and with some magic numbers. Alas, that’s the price to pay for orthogonal code!

PR #33, opened 2020-03-17
 📦 atlas-homepage-cards 💬 Bugfix #171843306 – Multiple carousels rendered on the same page don’t retain the prop `slidesToShow`

alfonsomunozpomer:
The original manifestation of the bug was that protists was rendered incorrectly in Single Cell Expression Atlas:
image

It turns out that for each carousel rendered in the same page the props were being mutated and inconsistencies arose. This was very tricky to find!

image


lingyun1010:
Since the carousel panel is implemented in the package `species-summary-panel` and we are using Lerna now, I think it might be good to add the `species-summary-panel` into this repository, as we cannot test the UI through this PR.

On the other hand, I tried to start the webpack server to check the bug in web-sc repository and the protists panel renders well with the old version carousel. Please have a look into the screenshot below.
Screenshot 2020-03-20 at 15 49 38

As we do not have protists data on local, I can not see the rendering results in JSP, which is pretty similar to the sc facet search results debugging situation. The local webpack servers run/render well, but not working on sc server.

alfonsomunozpomer:
@lingyun1010 Yes, the problem happens when you render multiple carousels, and the number of items to show differs. In the case from your screenshot, *Plants*, *Fungi* and *Protists* have all one species, so no problem. The issue manifests when a tab has more items than a previous tab. If you were to point at `wwwdev` to retrieve the species info you’ll see the bug. That’s why it was a bit hard to spot and track it down.

The issue in the facetted search I’d argue is quite different, since it’s a package version dependency, not about data.

In any case, yuo’re right about moving this package to the Lerna repo, but I’m hesitant about it since I don’t think we should publish that package. We can discuss how we can do it later.

PR #35, opened 2020-03-25
📦 scxa-gene-search-form 💬 Feature #171953862 – Add gene search form to CCA

alfonsomunozpomer:
For Cambridge Cell Atlas, having a text label on top of the text input doesn’t look very good:
image

So I added a prop so that we can hide it.

### DIFF
--- a/packages/scxa-gene-search-form/__test__/Autocomplete.test.js
+++ b/packages/scxa-gene-search-form/__test__/Autocomplete.test.js
@@ -80,6 +80,18 @@ describe(`Autocomplete suggestions fetch`, () => {
 
     expect(fetchMock.calls()).toHaveLength(1)
     expect(URI(fetchMock.lastUrl()).search(true)).toHaveProperty(`species`, species.join(`,`))
+  })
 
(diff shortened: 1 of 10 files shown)


packages/scxa-gene-search-form/__test__/Autocomplete.test.js
```diff
@@ -79,6 +80,18 @@ describe(`Autocomplete suggestions fetch`, () => {
...
+  })
+
+  test(`can hide the label if we pass an empty string`, () => {
+    const wrapper = shallow(<Autocomplete {...props} labelText={``} />)
+
+    expect(wrapper).not.toContainMatchingElement(`label`)
+  })
+
+  test(`can customise the label`, () => {
+    const labelText = `COVID-19 UK lockdown day 2`
```
haideriqbal: We will remember this day 😄

alfonsomunozpomer:
Yes, you’re right about the fetch loader, it’s on my TODO list. We’ll do it when me move this package to the monorepo.

PR #36, opened 2020-05-01
📦 atlas-experiment-table 💬 Bugfix #172189793 – Canis familiaris can’t find Canis lupus familiaris

alfonsomunozpomer:
The way to fix this issue was to broaden the default search results so that it matches any of the typed words in any order. In this way *Canis familiaris* matches *Canis lupus familiaris*, and also *familiaris canis* or *canis foobar* would. I decided to add an exact search if the searched phrase is enclosed in double quotes after running a poll in our Slack workspace.

### DIFF
--- a/packages/atlas-experiment-table/src/search.js
+++ b/packages/atlas-experiment-table/src/search.js
@@ -1,19 +1,32 @@
 import _ from 'lodash'
 
+// Don’t search for queries that are shorter than this
+const TOKEN_MIN_LENGTH = 3
+
 const fuzzyDeepContainsCaseInsensitive = (object, keys, value) => {
-  const words = _.chain(object)
-    .pick(keys)
-    .values()
-    .flatten()
-    .map(_.toString)
-    .map(_.toLower)
-    .map(phrase => _.split(phrase, /\s+/))
-    .flatten()
-    .value()
+  const words =
+    _.chain(object)
+      .pick(keys)
+      .values()
+      .flatten()
+      .map(_.toString)
+      .map(_.toLower)
+      .map(phrase => _.split(phrase, /\s+/))
+      .flatten()
+      .value()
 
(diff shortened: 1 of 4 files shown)


packages/atlas-experiment-table/src/search.js
```diff
@@ -0,0 +1,37 @@
...
+    .flatten()
+    .map(_.toString)
+    .map(_.toLower)
+    .some(valueToLower => valueToLower.includes(_.toString(value).toLowerCase()))
+    .value()
+
+export default (object, keys, value) => {
+  const trimmedValue = value.trim()
+
+  if (trimmedValue.charAt(0) === `"` && trimmedValue.charAt(trimmedValue.length - 1) === `"`) {
```
haideriqbal, line 32: 🎖️

alfonsomunozpomer:
Wow! I messed up big here. Let me have a look. :(

packages/atlas-experiment-table/src/search.js
```diff
@@ -0,0 +1,52 @@
+import _ from 'lodash'
+
+// Don’t search for queries that are shorter than this
+const TOKEN_MIN_LENGTH = 3
```
haideriqbal, line 4: Any reason why we have specifically chosen 3 as a number for search length? I personally prefer the way the functionality is now at the moment that as soon as I type a single character is searches for that.

alfonsomunozpomer: Usually one or two letter search queries mean very little. One letter is definitely meaningless, since we’re matching any word that will contain that letter. Two-letter words are arguably similar. But three letters is a lot more discriminatory. For example Solr does this by default, and I think it’s in general a good thing. We can change it if users become confused.

haideriqbal: Okay. This makes sense.

PR #45, opened 2020-06-12
Scxa tsne plot/bugfix/172954811 export as image fails

upendrakumbham:
Bug: When user try to download large experiments in the single-cell atlas, high charts export server giving error `

> 413 Request Entity Too Large

As per @alfonsomunozpomer  - added 'offline-exporting' module and wrap that in the high charts module which we are already doing. 

I have followed this [JSFiddle](https://jsfiddle.net/gh/get/jquery/1.11.0/highslide-software/highcharts.com/tree/master/samples/highcharts/exporting/offline-download-demo/) to fix this

### DIFF
--- a/packages/scxa-tsne-plot/html/Demo.js
+++ b/packages/scxa-tsne-plot/html/Demo.js
@@ -58,6 +58,21 @@ const experiment4 = {
   ]
 }
 
+//Image export fails for the large experiments
+// Number of cells: 32,369
+const experiment5 = {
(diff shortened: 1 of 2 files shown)


lingyun1010:
Good and clean implementation! Please check tests for this component by command ```npm test```.  We also have snapshots tests for the front-end component. You may find tests failed due to staled snapshots. You could update them by command ```npm test -- -u```.

PR #48, opened 2020-07-29
Pass sc host url to marker gene hyper link

lingyun1010:


packages/scxa-marker-gene-heatmap/src/MarkerGeneHeatmap.js
```diff
@@ -172,7 +172,7 @@ const MarkerGeneHeatmap = (props) => {
       labels: {
         useHTML: true,
         formatter: function () {
-          return `<a href="https://www.ebi.ac.uk/gxa/sc/search?q=${this.value}&species=${species}"` +
+          return `<a href="${host}search?q=${this.value}&species=${species}"` +
```
alfonsomunozpomer, line 175: Please, use a library to build links. Remember that we use `urijs` across our components for that.

PR #62, opened 2020-08-17
📦 scxa-tsne-plot 💬 Bugfix #174357060 – Not Available points not visible

alfonsomunozpomer:
Adding both an single colour point and also a series colour (seemingly) made Not Available points invisible. It’s either an undefined behaviour in Highcharts or the opacity was reduced twice (set to 0.65 × 0.65). With the added inconvenience that it wasn’t consistent, so I suspect it’s the former.


lingyun1010:
I don't really understand the logic behind it. It seems the colour property has not been changed, but why series instead of points solve the bug?

lingyun1010:
One notice that your commits are not verified.

alfonsomunozpomer:
Yes, I need to look into the GPG signature. I renewed it but maybe I need to upload it again.

Regarding the bug, the reason of why this happens and why this solves it, is in the opening PR post. I admit there’s a bit of speculation, it’s an educated guess.

PR #69, opened 2020-09-12
📦 scxa-tsne-plot 💬 Bugfix #174786132 – 0 expression values breaks the plot when integrated in the experiment page

alfonsomunozpomer:
**Very annoyingly**, this doesn’t manifest in the `scxa-tsne-plot` package, only when it’s integrated in the experiment page. Very probably, something to do with how rendering and initialising the Highcharts object is handled. The fact that [the updated negative log fiddle](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/coloraxis/logarithmic-with-emulate-negative-values/) changes the first event from `init` to `afterInit` reinforces my suspicion.

### DIFF
--- a/packages/scxa-tsne-plot/package-lock.json
+++ b/packages/scxa-tsne-plot/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@ebi-gene-expression-group/scxa-tsne-plot",
-  "version": "5.2.1",
+  "version": "5.3.0-alpha.0",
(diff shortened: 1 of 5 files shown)


lingyun1010 (approved):
Please add the jsfiddle link as a reference. LGTM.

alfonsomunozpomer:
The fiddle link is in the comments at the beginning of the module.

PR #70, opened 2020-09-17
Add Gene Ids dropdown in TSne Widget

haideriqbal:
Added functionality to pass gene ids as props in TSne Widget. Changes have also been made in TSne Plot view component as TSne widget uses that component to render TSne plot. In `GeneExpressionTSnePlot.js` changes have been to show dropdown if there is an array of gene ids otherwise it will show `AtlasAutocomplete` control by default.


lingyun1010 (changes requested):
If you test with some `geneIds` and select one of them to update the gene expression TSNE plot. You will find a bug in the props function `onSelectGeneId`. As the state `geneId` required as a string, but it was an object.

packages/scxa-tsne-plot/src/GeneExpressionTSnePlot.js
```diff
@@ -129,59 +130,78 @@ const GeneExpressionScatterPlot = (props) => {
...
+          enabled: true
+        }
+  }
+
+  const geneIdOptions = geneIds.sort((a, b) => a - b).map((geneId) => ({
+    value: geneId,
+    label: geneId
+  }))
+
+  let geneIdControl
```
alfonsomunozpomer, line 166: Better to use `const` and the ternary operator to do conditional rendering. There’s nothing essentially wrong with the above, but we use the latter pattern elsewhere on our codebase.

haideriqbal:
> If you test with some `geneIds` and select one of them to update the gene expression TSNE plot. You will find a bug in the props function `onSelectGeneId`. As the state `geneId` required as a string, but it was an object.

@lingyun1010 can you please refer me to the test where you are getting this error?

`geneId` being returned is indeed an object of shape `value: ..., label: ... ` so to assign it to the state variable I am doing this 
`onSelectGeneId={(geneId) => this.setState({ selectedGeneId: geneId.value })}`

Please let me know if I understood your query correctly or am I missing something?

lingyun1010 (approved):
LGTM

lingyun1010 (commented):
Please have a look into the failed tests (snapshots) in `scxa-tsne-widget` component.

lingyun1010:
Maybe we pass a prop `showControlls` instead of hard coding false if the widget component doesn’t need the dropdown at all. Or we could set a default prop value as `false` if you don’t want to bother the widget users.

haideriqbal:
You will see that snapshot test in `TSneWidget` has been commented out because it has some weird dependency with `react-select` being used in TSnePlotView` which I and @alfonsomunozpomer have agreed needs further investigation.

So for meanwhile I have commented out the test and implemented the suggested changes.

lingyun1010 (commented):
Thanks for the updates. A few general comments.
1. We prefer to have more tiny commits with proper commit subjects rather a merged commits.
2. Instead of having two `html` files, we have two options to do it properly in my opinion.
- If one `html` files just adds some the functionality (as we have in this PR), we use the comprehensive one as the default.
- If each `html` file has different functionality or views, like the [homepage-cards](https://github.com/ebi-gene-expression-group/atlas-components/blob/master/packages/atlas-homepage-cards/html/index.html) component, we display them all in the default `html` page or provide a way to switch between each other rather than changing the file name from `index1.html` to `index.html` manually.

lingyun1010 (approved):
LGTM！

PR #74, opened 2021-01-19
Feature #175872307 – Generalise to render cluster marker genes

alfonsomunozpomer:
Since the marker genes payload format changed singificantly, our old `scxa-marker-gene-heatmap` couldn’t render the clusters heatmap. In the end I think it was easier to take the cell types heatmap and inject a different title, tooltip, no data message and labels formatter. To achieve that the heatmap requires an extra prop, `heatmapType` that can take one of the two values from the provider object.

There are other minor corrections here and there.


lingyun1010 (commented):
Looks good but the marker gene hyperlink is hard-coded to `www.ebi.ac.uk` rather than accepting the `host` props.

lingyun1010 (approved):
All good to merge!

PR #76, opened 2021-02-18
New component: anatomogramExperimentTable

mvonpapen:
# Description
Based on Haider's and Lingyun's work we developed a new component: anatomogramExperimentTable

The component works correctly in the sense that selections on the anatomogram generate a list of associated experiments. However, it seems that only one experiment is associated to a specific part of the lung, namely to the "middle lobe of right lung" (UBERON_0002174). Selection of any other region results in an empty list of experiments.

image

# Todo
There are some things left to do before you can test it at EBI:
* in `package.json` line 43 a local anatomogram is loaded: `"@ebi-gene-expression-group/organ-anatomogram": "file:../../../organ-anatomogram"`
* in `anatomogramExperimentTable.js` you should check if the host (line 220) and the endpoint (line 39) are defined correctly
* in `index.html` check if URLs are defined correctly (lines 33-40)

# Issues
Clicking on a new region (or deselcting the selected region) in the anatomogram correctly leads to an update of `this.state.experiments`, which is consumed by experimentaTable. However, the experimentTable does not update automatically after a change of `this.state.experiments` - only after a click on any button within experimentTable (e.g. "Load date" or "Sort").

anatomoExpTable_bug


haideriqbal:
@mvonpapen after reviewing this PR, I and @alfonsomunozpomer have the following comments:

- In the HTML demo file, we should use `anatomogramHost: `http://localhost:9000/`` rather than just `localhost:9000/` as doing so gives CORS browser security error. So we need to give full URL.
- We have found the issue with Experiment Table not updating until some filter is clicked on. It is related to props being updating but not rerendering the experiment table. We are looking into two possible solutions in this regards that is either using React `ComponentDidUpdate` or `getDerivedStateFromProps`
- One thing we noticed in the code was use of jQuery for fetching the response. We recommend using fetch API instead. For inspiration, please have a look at [`atlas-fetch-loader`.](https://github.com/ebi-gene-expression-group/atlas-components/tree/master/packages/atlas-react-fetch-loader)

mvonpapen:
closed because PR moved to https://github.com/ebi-gene-expression-group/atlas-components/pull/98

PR #80, opened 2021-05-07
Re-render the component based on prop change

haideriqbal:
We had an issue while integrating Anatomograms and Experiment Table when a prop change was triggered by Anatomogram filtering. The prop change did not re-render the experiment table as there was no state change in the component. To fix this issue, I have implemented the componentDidUpdate function of React component lifecycle.


alfonsomunozpomer (approved):
LGTM. Did you test this against the DiscovAir component?

haideriqbal:
Yes, I did test it on DiscovAir component and it was working fine. I will now create a new release and merge it :)

PR #82, opened 2021-06-04
Scxa tsne plot/feature/178242718 refactor t sne with coords table and umap json

lingyun1010:
Please run the local sc server currently for getting responses with the latest tSNE JSON endpoints.
The corresponding PRs are [web-core](https://github.com/ebi-gene-expression-group/atlas-web-core/tree/feature%2F174839971-add-projection-method-to-tSNE) and [web-single-cell](https://github.com/ebi-gene-expression-group/atlas-web-single-cell/tree/feature%2F174839971-add-projection-method-to-tSNE).

### DIFF
--- a/packages/scxa-tsne-plot/__test__/ClusterTSnePlot.test.js
+++ b/packages/scxa-tsne-plot/__test__/ClusterTSnePlot.test.js
@@ -94,16 +94,18 @@ describe(`ClusterTSnePlot`, () => {
     },
     plotTypeDropdown: [
       {
-        plotType: `TSne`,
+        plotType: `tSNE`,
         plotOptionsLabel: `Perplexities`,
         plotOptions: [1, 2, 3, 4]
       },
       {
-        plotType: `UMap`,
+        plotType: `UMAP`,
         plotOptionsLabel: `N-neighbors`,
         plotOptions: [1, 2, 3]
       }
     ],
+    selectedParameter: 10,
+    selectedPlotType: `tSNE`,
     height: 500,
     metadata: [],
     ks: [],
@@ -113,26 +115,26 @@ describe(`ClusterTSnePlot`, () => {
 
   test(`with no data matches snapshot`, () => {
     const tree = renderer
-      .create(<ClusterTSnePlot {...props} selectedColourBy={`0`} selectedPerplexity={0} showControls={true}/>)
+      .create(<ClusterTSnePlot {...props} selectedColourBy={`0`} showControls={true}/>)
       .toJSON()
 
     expect(tree).toMatchSnapshot()
   })
 
   test(`contains ScatterPlotLoader`, () => {
-    const wrapper = mount(<ClusterTSnePlot {...props} selectedColourBy={`0`} selectedPerplexity={0} showControls={true}/>)
+    const wrapper = mount(<ClusterTSnePlot {...props} selectedColourBy={`0`} showControls={true}/>)
 
     expect(wrapper.find(ScatterPlotLoader).length).toBe(1)
   })
 
   test(`contains 3 PlotSettingsDropdowns`, () => {
-    const wrapper = mount(<ClusterTSnePlot {...props} selectedColourBy={`0`} selectedPerplexity={0} showControls={true}/>)
+    const wrapper = mount(<ClusterTSnePlot {...props} selectedColourBy={`0`} showControls={true}/>)
 
-    expect(wrapper.find(PlotSettingsDropdown).length).toBe(3)
+    expect(wrapper.find(PlotSettingsDropdown).length).toBe(1)
   })
 
   test(`contains only 1 PlotSettingsDropdowns when showControls is false`, () => {
-    const wrapper = mount(<ClusterTSnePlot {...props} selectedColourBy={`0`} selectedPerplexity={0} showControls={false}/>)
+    const wrapper = mount(<ClusterTSnePlot {...props} selectedColourBy={`0`} showControls={false}/>)
 
     expect(wrapper.find(PlotSettingsDropdown).length).toBe(1)
   })
@@ -146,7 +148,7 @@ describe(`ClusterTSnePlot`, () => {
       }
     ]
 
-    const wrapper = mount(<ClusterTSnePlot {...props} ks={ks} metadata={metadata} selectedColourBy={`2`} selectedPerplexity={0} showControls={true}/>)
+    const wrapper = mount(<ClusterTSnePlot {...props} ks={ks} metadata={metadata} selectedColourBy={`2`} showControls={true}/>)
 
     const dropdown = wrapper.find({ labelText: `Colour plot by:`})
 
@@ -163,7 +165,7 @@ describe(`ClusterTSnePlot`, () => {
       }
     ]
 
-    const wrapper = mount(<ClusterTSnePlot {...props} ks={ks} metadata={metadata} selectedColourBy={`metadata-1`} selectedPerplexity={0} showControls={true}/>)
+    const wrapper = mount(<ClusterTSnePlot {...props} ks={ks} metadata={metadata} selectedColourBy={`metadata-1`} showControls={true}/>)
 
     const dropdown = wrapper.find({ labelText: `Colour plot by:`})
 
--- a/packages/scxa-tsne-plot/html/Demo.js
+++ b/packages/scxa-tsne-plot/html/Demo.js
@@ -132,7 +132,7 @@ class Demo extends React.Component {
     this.state = {
       selectedPlotType: plotTypeDropdown[0].plotType.toLowerCase(),
       geneId: ``,
-      selectedParameter: plotTypeDropdown[0].plotOptions[0],
+      selectedPlotOption: plotTypeDropdown[0].plotOptions[0],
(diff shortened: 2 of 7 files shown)


packages/scxa-tsne-plot/html/Demo.js
```diff
@@ -172,34 +174,48 @@ class Demo extends React.Component {
...
           </form>
         </div>
+   */
 
+  render() {
+    return(
+       <div className={`row column expanded`}>
         <TsnePlotView
-          atlasUrl={`https://wwwdev.ebi.ac.uk/gxa/sc/`}
+          atlasUrl={`http://localhost:8080/gxa/sc/`}
```
haideriqbal, line 183: Please change the localhost URL to public or dev URL, please. So that someone running the demo in the future doesn't have to run anything on their local machine.

packages/scxa-tsne-plot/src/ClusterTSnePlot.js
```diff
@@ -263,10 +203,8 @@ ClusterTSnePlot.propTypes = {
   onChangeColourBy: PropTypes.func,
   clusterType: PropTypes.string,
 
-  perplexities: PropTypes.arrayOf(PropTypes.number).isRequired,
-  selectedPerplexity: PropTypes.number.isRequired,
-  onChangePerplexity: PropTypes.func.isRequired,
-
+  selectedPlotType: PropTypes.string.isRequired,
+  selectedParameter: PropTypes.number.isRequired,
```
haideriqbal, line 207: These props are marked as required but are not being used in the component.

packages/scxa-tsne-plot/src/ClusterTSnePlot.js
```diff
@@ -263,10 +203,8 @@ ClusterTSnePlot.propTypes = {
...
   onChangePlotOptions: PropTypes.func.isRequired,
   onChangePlotTypes: PropTypes.func.isRequired,
```
haideriqbal, line 209: These props are not being used either so it would be good to remove them if they are not required.

packages/scxa-tsne-plot/src/GeneExpressionTSnePlot.js
```diff
@@ -165,7 +165,8 @@ const GeneExpressionScatterPlot = (props) => {
 
 
   const geneIdControl = geneIds.length ? <PlotSettingsDropdown
-          labelText={`Gene Ids`}
+          labelText={`Gene Ids:`}
+          labelText={`Gene Ids:`}
```
haideriqbal, line 169: Please remove this. I assume the extra statement has been added by mistake.

packages/scxa-tsne-plot/src/TSnePlotView.js
```diff
@@ -119,26 +128,47 @@ class TSnePlotView extends React.Component {
...
+                onSelect={(selectedPlotType) => {
+                  onChangePlotTypes(selectedPlotType.value)
+                }}/>
+          </div>
+          <div className={`small-12 medium-6 columns`}>
+            <PlotSettingsDropdown
+                labelText={plotOptionsLabel}
+                options={plotOptionsValues}
+                defaultValue={{value: selectedParameter,
+                  label: selectedPlotType == `tSNE` ? `perplexity=` + selectedParameter : `n_neighbor=` + selectedParameter}}
```
haideriqbal, line 146: Please replace `==` with `===` as `==` is a loose equality operator in javascript.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness

packages/scxa-tsne-plot/src/TSnePlotView.js
```diff
@@ -176,19 +206,17 @@ TSnePlotView.propTypes = {
   wrapperClassName: PropTypes.string,
   clusterPlotClassName: PropTypes.string,
   expressionPlotClassName: PropTypes.string,
+  selectedParameter: PropTypes.number.isRequired,
```
haideriqbal, line 209: What do you think if we rename `selectedParameter` to `selectedPlotOption`. In my opinion, it would be more readable and easy to understand.

haideriqbal:
At the moment the plot option label is not changing corresponding to the plot type. The label remains as `n-neighbor` no matter if it's tSNE or UMAP.

Screenshot 2021-06-04 at 08 26 31

I am not sure if this was discussed in the team meeting while I was on leave so please bear with me if I am wrong. I see all 4 dropdowns evenly spread out on top but I feel that putting the `Options` dropdown above the `gene expression` plot might confuse the users that it's impacting the gene-expression plot where in reality it isn't.

Screenshot 2021-06-04 at 08 54 21

haideriqbal (changes requested):
I have left some comments please review. It's mostly related to removing unnecessary bits of code and one minor bug.

packages/scxa-tsne-plot/__test__/ClusterTSnePlot.test.js
```diff
@@ -94,16 +94,18 @@ describe(`ClusterTSnePlot`, () => {
     },
     plotTypeDropdown: [
       {
-        plotType: `TSne`,
+        plotType: `tSNE`,
```
upendrakumbham, line 97: A better approach in my view, load the backend response dynamically into `Plotype` and `Options` dropdowns without changing much. So that in the future, if we added more projection methods and its options, no need to change the front-end code.

packages/scxa-tsne-plot/html/Demo.js
```diff
@@ -93,12 +93,10 @@ const experiment6 = {
   plotTypeDropdown: [
     {
       plotType: `UMAP`,
-      plotOptionsLabel: `Options:`,
       plotOptions: [5, 10, 15, 20, 25, 50, 100]
```
upendrakumbham: `UMAP options:`

- n_neighbors": 5
- n_neighbors": 100..etc

packages/scxa-tsne-plot/html/Demo.js
```diff
@@ -93,12 +93,10 @@ const experiment6 = {
...
     },
     {
       plotType: `tSNE`,
```
upendrakumbham: tSNE options:

-  perplexity : 40
-  perplexity: 25

upendrakumbham: But I could see only integers

upendrakumbham: As this is a demo page, we can use anything I guess.

packages/scxa-tsne-plot/html/Demo.js
```diff
@@ -90,16 +90,11 @@ const experiment6 = {
...
-    {
-      plotType: `tSNE`,
-      plotOptions: [1, 5 ,10, 15, 20, 25, 30, 35, 40, 45, 50]
-    }
-  ]
+  plotTypesAndOptions: {
+    "tsne": [{ "perplexity": 40 }, { "perplexity": 25 }, { "perplexity": 45 },{ "perplexity": 1 },{ "perplexity": 30 },
+    {"perplexity": 10 },{ "perplexity": 15 },{ "perplexity": 50 },{ "perplexity": 35 },{ "perplexity": 20 },{ "perplexity": 5 }],
+    "umap": [{"n_neighbors": 5},{"n_neighbors": 100},{"n_neighbors": 50},{"n_neighbors": 10},{"n_neighbors": 30},{"n_neighbors": 15},{"n_neighbors": 3}]
+  }
```
upendrakumbham, line 97: This is perfect, please ignore the above comments on the demo page.

packages/scxa-tsne-plot/html/Demo.js
```diff
@@ -132,7 +138,8 @@ class Demo extends React.Component {
     this.state = {
       selectedPlotType: plotTypeDropdown[0].plotType.toLowerCase(),
       geneId: ``,
-      selectedPlotOption: plotTypeDropdown[0].plotOptions[0],
+      selectedPlotOption: Object.values(plotTypeDropdown[0].plotOptions[0])[0],
+      selectedPlotOptionLabel: Object.keys(plotTypeDropdown[0].plotOptions[0])[0] + `: ` + Object.values(plotTypeDropdown[0].plotOptions[0])[0],
       selectedColourBy: ks[Math.round((ks.length -1) / 2)].toString(),
```
upendrakumbham, line 143: is this calculation for the preferred `Options`  dropdown value while loading the page for the first time?

packages/scxa-tsne-plot/html/Demo.js
upendrakumbham: ignore

packages/scxa-tsne-plot/html/Demo.js
upendrakumbham: I ran the branch in my local, I have a question for @lingyun1010 :

1. When I change to `tSNE` plotType - corresponding preferred value is : `perplexity:40`
image

2. When I change 'UMAP' plotType - corresponding preferred value is: `n_neighbors:5`

image

is this the correct preferred value for the `UMAP` Options?

upendrakumbham: I and @lingyun1010  had a meeting to discuss review comments, which got clarified.

upendrakumbham (approved):
LGTM!

PR #84, opened 2021-07-26
Make inferred cell type as default if present

haideriqbal:
This is an old feature that somehow got removed so reverted back the relevant changes, here is the older PR https://github.com/ebi-gene-expression-group/atlas-components/pull/72. Issue explanation: https://github.com/ebi-gene-expression-group/atlas-web-single-cell/issues/181.

### DIFF
--- a/packages/scxa-tsne-plot/__test__/ClusterTSnePlot.test.js
+++ b/packages/scxa-tsne-plot/__test__/ClusterTSnePlot.test.js
@@ -173,7 +173,7 @@ describe(`ClusterTSnePlot`, () => {
     expect(dropdown.props().defaultValue.label).toContain(`The first metadata value`)
   })
 
-  test(`dropdown selection text is inferred_cell_type when metadata options contain when plot is coloured by metadata`, () => {
+  test(`dropdown selection text is Inferred cell type - authors_labels when plot is coloured by metadata`, () => {
     const ks = [1, 2, 3]
     const metadata = [
       {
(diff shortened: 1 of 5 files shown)


packages/scxa-tsne-plot/__test__/ClusterTSnePlot.test.js
```diff
@@ -172,6 +173,26 @@ describe(`ClusterTSnePlot`, () => {
     expect(dropdown.props().defaultValue.label).toContain(`The first metadata value`)
   })
 
+  test(`dropdown selection text is inferred_cell_type when metadata options contain when plot is coloured by metadata`, () => {
```
alfonsomunozpomer, line 176: Fix grammar, please.

alfonsomunozpomer (changes requested):
In order to be consistent with the cell type wheel we should use the following metadata fields in order:
1. Inferred cell type - ontology labels
2. Inferred cell type - authors labels
3. Cell type
4. Progenitor cell type

alfonsomunozpomer (changes requested):
The proper value is selected in the dropdown, but notice that the left plot doesn’t honour the setting in the initial request, which is (e.g. in the demo page):
`https://wwwdev.ebi.ac.uk/gxa/sc/json/experiments/E-MTAB-5061/tsneplot/5/clusters/variable/20?method=umap&variableType=clusters`

lingyun1010 (changes requested):
The dropdown box is not responding if you select other options in the list.

haideriqbal:
Closing this PR as this bug was resolved in other PR on backend here https://github.com/ebi-gene-expression-group/atlas-web-single-cell/pull/186

PR #85, opened 2021-07-27
Feature #172328184 – Cell Type Wheel demo

alfonsomunozpomer:
Please test the demos with `npx webpack serve --mode=development`.

This is the bare minimum component. I’m aware that you worked on tps://github.com/ebi-gene-expression-group/atlas-cell-type-wheel/pull/7/ and that’s still useful, but React, Highcharts and Webpack have significantly changed since then, and we need to adopt new testing strategies, so let’s rebuild this with the bare minimum. We can pick and apply changes from your original wheel to reconcile the JSON schema and make it a bit more feature rich, one step at a time.

Tests will for the time being fail, that will be the next thing.


packages/scxa-cell-type-wheel/src/CellTypeWheel.js
```diff
@@ -0,0 +1,85 @@
...
+// SUPER HACK!
+// https://github.com/highcharts/highcharts/issues/4994
+window.Highcharts = Highcharts
+
+class CellTypeWheel extends React.Component {
+  constructor(props) {
+    super(props)
+
+    this.state = {
+      // To avoid unnecessary update keep all options in the state.
```
lingyun1010, line 17: Sorry I did not get this comment, whey do we need to set those options as state? I think we could utilize a stateless component here.

lingyun1010, line 17: Sorry I did not get this comment, whey do we need to set those options as state? I think we could utilize a stateless component here.

alfonsomunozpomer: Sorry, this was a bit of a distracting comment. I copy-pasted the template from https://github.com/highcharts/highcharts-react#optimal-way-to-update, and it’s actually an explanation of the Highcharts wrapper, not about our code. I’ve removed it.

lingyun1010: Yes, but still we do not need a stateful component here as far as I can see.

alfonsomunozpomer: In principle we don’t, but observer that ultimately the `render` is called by `HighchartsReact` and you don’t know how it manages things internally. Highcharts official documentation states that:
> A good practice is to keep all chart options in the state. When `setState` is called, the options are overwritten and only the new ones are passed to the `chart.update` method.

I say, let’s go with that and we’ll have fewer surprises down the road. Maybe everything’s fine with a purely functional component, but I prefer to stick to Highcharts docs so if we get some unexepected behaviour we can report it (and get an upstream fix) sooner.

lingyun1010: Okay, no problem. I did not know that Highcharts suggests in this way. I think functional component is lighter than stateful ones.

lingyun1010 (approved):
LGTM!

PR #86, opened 2021-07-28
Grey ou the not defined points

lingyun1010:
Regarding the issue [Not defined labels are not handled as Not available as it used to be](https://github.com/ebi-gene-expression-group/atlas-web-single-cell/issues/182)


pcm32:
I think that in the previous functionality we transformed `not defined` points into `Not available`, instead of keeping the two sets of points separate.

pcm32:
But I'm not keen on seen two different labels now, `Not defined` and `Not available.`

PR #89, opened 2021-08-27
Add accessKey prop

haideriqbal:
Cell plot data urls were querying backend without accessKey param so to make them work for private experiments we have to add accessKey params to the URLs so that they are able to retrieve private experiments.

### DIFF
--- a/packages/scxa-tsne-plot/__test__/TSnePlotView.test.js
+++ b/packages/scxa-tsne-plot/__test__/TSnePlotView.test.js
@@ -101,7 +101,7 @@ describe(`TSnePlotView`, () => {
     expect(wrapper.state(`cellClustersData`)).toEqual(payload)
   })
 
-  test(`does not updates its state with the JSON payload after a unsuccessful request`, async () => {
+  test(`does not update its state with the JSON payload after an unsuccessful request`, async () => {
     const payload = {
       series: [`foo`, `bar`, `horse`]
     }


upendrakumbham (commented):
LGTM!. But any changes are required for the unit tests as we added a new prop?

haideriqbal:
> LGTM!. But any changes are required for the unit tests as we added a new prop?

Thanks, Upendra for reviewing. I believe the only thing changing as a result of this feature is adding an extra param for cell plot data URL. We had the URL before as well and there are no tests for `TSnePlotView.js` and data being fetched is ultimately tested in other classes downstream. So, in my opinion, we are good without a test in this situation.

upendrakumbham (approved):
Thanks for the explanation!. LGTM!

packages/scxa-tsne-plot/src/TSnePlotView.js
```diff
@@ -114,12 +116,12 @@ class TSnePlotView extends React.Component {
...
     const plotTypesOptions = plotTypeDropdown.map((plot) => ({
-          value: plot.plotType.toLowerCase(),
-          label: plot.plotType
-        }))
+      value: plot.plotType.toLowerCase(),
+      label: plot.plotType
+    }))
 
     const getTooltipContent = async (cellId) => {
       const url = URI(`json/experiment/${this.props.experimentAccession}/cell/${cellId}/metadata`, atlasUrl).toString()
```
alfonsomunozpomer, line 127: The `accessKey` should be propagated to this request as well. Remember: if there’s an `experimentAccession` involved, then we should honour the `accessKey` field. Has the controller for this URL in the back end been modified?

alfonsomunozpomer (changes requested):
There’s one request which lacks `accessKey` that should be added.

alfonsomunozpomer:
> Thanks for the explanation!. LGTM!

@upendrakumbham Haider’s explanation didn’t convince me, so don’t let him get away with it! @haideriqbal I think there’s a good test hiding in the shadows.

packages/scxa-tsne-plot/src/TSnePlotView.js
haideriqbal: The controller already implements the logic correctly using accessKey so change in the frontend would suffice.

packages/scxa-tsne-plot/__test__/TSnePlotView.test.js
```diff
@@ -0,0 +1,100 @@
...
+        experimentAccession={accession}
+        selectedPlotOption={5}
+        plotTypeDropdown={plotTypeDropdown}
+        selectedPlotOptionLabel={``}
+      />)
+      .toJSON()
+    expect(tree).toMatchSnapshot()
+  })
+
+  test(`whether accessKey prop is passed or not`, async () => {
```
alfonsomunozpomer, line 71: This test name should be improved. The test name should describe your expectation. For instance, here you should say something like: “updates its state with the JSON payload after a successful request”. While this test is useful, it can be complemented with another test that verifies that wihtout the `accessKey` prop the state isn’t updated.

If, on the other hand (or addionally), you want to test that “the accessKey prop is used as a query parameter” you can use [fetch-mock’s inspection API](https://github.com/wheresrhys/fetch-mock/blob/master/docs/cheatsheet.md#inspecting-calls) and verify that the request includes `accessKey`.

packages/scxa-tsne-plot/__test__/TSnePlotView.test.js
```diff
@@ -92,9 +93,74 @@ describe(`TSnePlotView`, () => {
...
 
+    expect(wrapper.state(`loadingCellClusters`)).toEqual(true)
+
     await wrapper.instance().componentDidMount()
+    await fetchMock.flush(true) // necessary to wait for all the fetch promises to be resolved before making any assertions
 
     expect(wrapper.state(`cellClustersData`)).toEqual(payload)
   })
 
+  test(`does not updates its state with the JSON payload after a unsuccessful request`, async () => {
```
alfonsomunozpomer, line 104: Two typos:
- “does not update”
- “an unsuccessful”

alfonsomunozpomer (approved):
Fix two typos and done!

PR #90, opened 2021-09-24
Chore #178281433 / Feature #178622182 – Cell Type Wheel and Cell Type Heatmap

alfonsomunozpomer:
I added [the view feature](https://www.pivotaltracker.com/story/show/178622182) since the changes in this branch overlap somewhat with that story. I think  we can start with this and then wire the two components later.

@haideriqbal Let’s publish the cell type heatmap as 2.1.0 (not beta), and the other two components as 1.0.0-beta as we discussed. The changes in the heatmap are backwards compatible.


packages/scxa-cell-type-wheel-experiment-heatmap/src/CellTypeWheelExperimentHeatmap.js
```diff
@@ -0,0 +1,59 @@
...
+import MarkerGeneHeatmap from '@ebi-gene-expression-group/scxa-marker-gene-heatmap/lib/MarkerGeneHeatmap'
+import CellTypeWheel from '@ebi-gene-expression-group/scxa-cell-type-wheel'
+
+class CellTypeWheelExperimentHeatmap extends React.Component {
+  constructor(props) {
+    super(props)
+  }
+
+  render() {
+    const { cellTypeWheelData, cellTypeWheelSearchTerm, cellTypeHeatmapSearchTerm, cellTypeHeatmapData } = this.props
```
haideriqbal, line 15: `cellTypeHeatmapSearchTerm` and `cellTypeHeatmapData` are not being passed in props. Please add these two in props as well.

packages/scxa-cell-type-wheel-experiment-heatmap/src/CellTypeWheelExperimentHeatmap.js
```diff
@@ -0,0 +1,59 @@
...
+            heatmapType={`celltypes`}
+          />
+        </div>
+      </div>
+    )
+  }
+}
+
+CellTypeWheelExperimentHeatmap.propTypes = {
+  // cellTypeHeatmapData: PropTypes.arrayOf(PropTypes.shape({
```
haideriqbal, line 46: we would need to change this in the format expected by `MarkerGenesHeatmap`. Previously it was formatted to match the Heatmap data

haideriqbal (changes requested):
One minor thing regarding props in the new component. Otherwise looks good!

haideriqbal (approved):
LGTM!

PR #93, opened 2021-10-13
Format cell type wheel tooltip

haideriqbal:
Very small change to format the tooltip by removing the value in the tooltip


alfonsomunozpomer (approved):
Simple and good.

PR #94, opened 2021-10-14
Elipsify titles heatmap celltype wheel

haideriqbal:
Ellipsis the titles of marker gene heatmap and cell type wheel. Highchart has no inbuilt functionality to add a tooltip to the title so had to do event handling to achieve tooltip functionality on the title.

### DIFF
--- a/packages/scxa-cell-type-wheel/html/leukocyte.html
+++ b/packages/scxa-cell-type-wheel/html/leukocyte.html
@@ -32,7 +32,7 @@
   <script src="../dist/myPackageDemo.bundle.js"></script>
 
   <script>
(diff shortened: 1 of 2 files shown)


packages/scxa-cell-type-wheel/html/leukocyte.html
```diff
@@ -32,7 +32,7 @@
   <script src="../dist/myPackageDemo.bundle.js"></script>
 
   <script>
-    myPackageDemo.render({}, 'leukocyte', 'target')
+    myPackageDemo.render({}, 'leukocyte', 'leukocyte', 'target')
```
alfonsomunozpomer, line 35: Why do you need the two `'leukocyte'` here?

PR #98, opened 2022-02-17
Implement new component: anatomogramExperimentTable

mazurkieth:
Based on Haider's and Lingyun's work we developed a new component: `anatomogramExperimentTable`.

image

It allows you to filter the experiment table by selecting organ parts in the Anatomogram.
Furthermore it allows to add a description that is displayed to the right of the Anatomogram.

To retrieve the data we utilize the fetchloader.


packages/atlas-anatomogram-experiment-table/src/anatomogramExperimentTable.js
```diff
@@ -0,0 +1,313 @@
...
+  constructor(props) {
+    super(props)
+    this.state = {
+      selectedSpecies: props.species,
+      showIds: props.showIds,
+      experiments: props.experiments,
+      highlightIds: [],
+      selectIds: [],
+      selectAllIds: [],
+      selectedOrganId: ``
```
alfonsomunozpomer, line 27: Since we’re also changing the `resource` later via `setState`, let’s initialise it here for completeness:
```javascript
this.state = {
  ...,
  resource: props.resource
}
```

packages/atlas-anatomogram-experiment-table/src/anatomogramExperimentTable.js
```diff
@@ -0,0 +1,313 @@
...
+    } else {
+      this.setState({
+        selectIds: ids,
+        resource:
+          URI(`json/experiments/${this.props.experimentAccession}/marker-genes/cell-types`, this.props.host)
+            .search({ organismPart: ids.map(ontologyAccessionToOntologyUri).join(`,`) })
+            .toString()
+      })
+    }
+    this.getExperimentsAndUpdateState(this.props)
```
alfonsomunozpomer, line 60: Since `getExperimentsAndUpdateState` always uses the host from `props` and the `resource` from the state, I think we can modify it so it doesn’t take any paramteres at all.

packages/atlas-anatomogram-experiment-table/src/anatomogramExperimentTable.js
```diff
@@ -0,0 +1,313 @@
...
+  }
+
+  async getExperimentsAndUpdateState({ resource, host }) {
+    this.setState({
+      isLoading: true
+    })
+
+    const url = URI(resource, host).toString()
```
alfonsomunozpomer, line 68: We can do `URI(this.state.resource, this.props.host)` to get the right URL.

packages/atlas-anatomogram-experiment-table/html/anatomogramExperimentTableDemo.js
```diff
@@ -0,0 +1,31 @@
...
+    .reduce((acc, svgMetadata) => acc.concat(svgMetadata.ids), [])
+    .filter(unique)
+    .sort()
+
+const render = (options, target) => {
+  console.log("options=", options);
+  ReactDOM.render(<AnatomogramExperimentTableWithFetchLoader
+    {...options}
+    host={options.dataHost}
+    resource={`json/experiments/hca/human/`}
```
alfonsomunozpomer, line 25: Let’s start with `json/experiments/hca/human?organismPart=${options.species}` to go with just lung experiments.

packages/atlas-anatomogram-experiment-table/src/anatomogramExperimentTable.js
```diff
@@ -0,0 +1,313 @@
...
+                  dataKey: `loadDate`,
+                  sortable: true,
+                  width: 0.5
+                },
+                {
+                  label: `Species`,
+                  dataKey: `species`,
+                  searchable: true,
+                  sortable: true
+                },
```
alfonsomunozpomer, line 234: Since we’re going with only human experiments I don’t know if we need this one. Definitely not `searchable` or `sortable`. Do we have plans of including non-human experiments?

packages/atlas-anatomogram-experiment-table/src/anatomogramExperimentTable.js
```diff
@@ -0,0 +1,313 @@
...
+                    return `experiments/` + dataRow.experimentAccession + `/experiment-design`
+                  },
+                  width: 0.5
+                }
+              ]}
+              dropdownFilters={[
+                {
+                  label: `Kingdom`,
+                  dataKey: `kingdom`
+                },
```
alfonsomunozpomer, line 264: Again, no need to choose kingdom, it’ll be only animals. :)

alfonsomunozpomer (changes requested):
Some comments to make the filtering work. We can have the functionality there with those changes.

packages/atlas-anatomogram-experiment-table/src/anatomogramExperimentTable.js
```diff
@@ -0,0 +1,305 @@
...
+    }
+    this.getExperimentsAndUpdateState()
+  }
+
+  async getExperimentsAndUpdateState() {
+    this.setState({
+      isLoading: true
+    })
+
+    const url = URI(this.props.resource, this.props.host).toString()
```
alfonsomunozpomer, line 64: This should be picking the resouce from the state, instead of props. Otherwise filtering by clickin on an organism part won’t be reflected in the table rows.

alfonsomunozpomer (changes requested):

packages/atlas-anatomogram-experiment-table/src/anatomogramExperimentTable.js
mazurkieth: Thank you for pointing this out, fixed that.

alfonsomunozpomer (approved):
I’ve noticed a couple of bugs, but at this point I think we’re in a good place to get the component added to the repo and work on individual issues.

PR #99, opened 2022-02-25
:package: scxa-marker-gene-heatmap :speech_balloon: Feature #179733229 – Show cell types as well as clusters

lingyun1010:
I modified the marker gene heatmap component for metadata plot options with data from a back-end PR https://github.com/ebi-gene-expression-group/atlas-web-single-cell/pull/223


alfonsomunozpomer (changes requested):
Do we need both `PlotSettingsDropdown.js` and `PlotSettingsDropdown2.js`? They are very similar! If you can’t get refactor both into a single component, could you please rename `PlotSettingsDropdown2.js` to something more meaningfu and add tests to document the requirements of each component?

packages/scxa-marker-gene-heatmap/src/clustersView/ClustersHeatmapView.js
```diff
@@ -77,9 +75,9 @@ class HeatmapView extends React.Component {
   }
 
   render() {
-    const { data, filteredData, selectedClusterId, isLoading, hasError } = this.state
+    const { data, filteredData, isLoading, hasError } = this.state
     const { wrapperClassName, plotWrapperClassName } = this.props
-    const { ks, ksWithMarkers, selectedK, onSelectK } = this.props
+    const { ks, ksWithMarkers, selectedColourBy, onChangeMarkerGeneFor, onChangeColourBy, metadata, cellTypes, selectedColourByCategory, selectedClusterId} = this.props
```
alfonsomunozpomer, line 80: The variable names `selectedColourBy`, `onChangeColourBy` and `selectedColourByCategory` were appropriate in the context of the cell plots because the scatter plot series would get different colours depending on these values. However, in the heatmap it’s not the case. Could you rename these variables to something better?

packages/scxa-marker-gene-heatmap/src/clustersView/ClustersHeatmapView.js
```diff
@@ -115,23 +140,24 @@ class HeatmapView extends React.Component {
           <div className={wrapperClassName}>
             <div className={`small-12 medium-6 columns`}>
               <PlotSettingsDropdown
-                labelText={`Number of clusters:`}
-                options={kOptions}
-                onSelect={(selectedOption) => onSelectK(selectedOption.value)}
-                value={{value: selectedK, label: `k = ${selectedK}`}}
-              />
+                  labelText={`Colour plot by:`}
```
alfonsomunozpomer, line 143: Again, we’re not colouring the plot. Change the text accordingly, please.

packages/scxa-marker-gene-heatmap/src/clustersView/ClustersHeatmapView.js
```diff
@@ -90,15 +88,42 @@ class HeatmapView extends React.Component {
...
+            options.map((item) => (item.options))
+        ),
+        {value: selectedColourBy}
+    )
+
+    const allClusterIds = selectedColourByCategory == `metadata` ? cellTypes :
+        _.range(1, parseInt(selectedColourBy) + 1)
+
+
+    //TODO
```
alfonsomunozpomer, line 118: Is there anything left to do? If not, remove this comment. :wink:

alfonsomunozpomer (changes requested):
Some variable names and displayed text should be changed. Also, notice that when a metadata field has been selected the “all” value shows the label “All clusters”, when that’s not correct. Almost there!

alfonsomunozpomer (changes requested):
The plot type in all cases is a heatmap, so the `plotType` in the variable names is redundant. Think what it is that you’re actually selecting.

Also, notice that the structure for the label is \<Verb\> + \<Object\> because it’s a command. The sentence “Metadata or cluster plot by” isn’t grammatically correct because “metadata” isn’t a verb.

alfonsomunozpomer (approved):
I think this is definitely a better name. Merge away!

PR #101, opened 2022-03-10
📦 Atlas Information Banner – 💬 Version 2

alfonsomunozpomer:
I revamped the component so that it can render a Markdown message. It leverages Foundation from the EBI Visual Framework, in particular the [Close Button](https://get.foundation/sites/docs/close-button.html). The reason for this change is to simplify the component and make it purely functional with no side effects.

It’s not opinionated about where and how to position the banner, giving the client a bit more flexibility (again, it should use [Foundation’s Float Grid](https://get.foundation/sites/docs/grid.html) as shown in the demo page).

## Tests
It’s worth noting that we can’t test that the component disappears after clicking on the closing button since that is delegated by Foundation to jQuery. Also, the package we’re using to render Markdown, [react-markdown](https://github.com/remarkjs/react-markdown), breaks Jest at the time of writing and we ultimately can’t test the correct transformation of the message. However, both of these issues aren’t really problematic since we shouldn’t test code maintained by others.

Of course react-markdown shouldn’t break *any* testing framework, but fortunately we can mock imports with Jest and output the message as raw text.

## Addendum
If for some reason we wish to control visibility ourselves, I leave this as reference. This version of the component is still purely functional thanks to the [`useState` hook](https://reactjs.org/docs/hooks-state.html).

```javascript
const AtlasInformationBanner = () => {
  const [visible, setVisible] = useState(true)

  return (typeof motd === `string` && motd.trim().length > 0) ?
    (
      <div className={`callout margin-bottom-none`} data-closable={``} role={`complementary`}>
        <ReactMarkdown remarkPlugins={[remarkGfm]} rehypePlugins={[rehypeRaw]}>{motd}</ReactMarkdown>
        <button onClick={() => setVisible(false) } style={{fontSize:`xx-large`, border: `none`}}>
          <span aria-hidden={true}>&times;</span>
        </button>
      </div>
    ) :
    null
}
```


packages/atlas-information-banner/src/AtlasInformationBanner.js
```diff
@@ -1,33 +1,31 @@
...
+    ) :
+    null
+}
 
-    this.state = {
-      visibility: ``
-    }
-  }
+AtlasInformationBanner.propTypes = {
+  motd: PropTypes.string
```
haideriqbal, line 24: motd means `Message of the Day` right?

haideriqbal (approved):
LGTM! One very minor comment. Please fix the file conflicts before merging.

packages/atlas-information-banner/src/AtlasInformationBanner.js
alfonsomunozpomer: Right. :)

alfonsomunozpomer:
> LGTM! One very minor comment. Please fix the file conflicts before merging.

No need to mention the latter, otherwise we wouldn’t be able to merge!

PR #102, opened 2022-03-10
📦 Atlas React Fetch Loader – 💬 Add ability to fetch raw text

alfonsomunozpomer:
Add optional raw prop to unpack the response as text (defaults to `false`). We want this so that we can combine it with the [latest iteration of the information banner](https://github.com/ebi-gene-expression-group/atlas-components/pull/101) and fetch a Markdown file from a URL and display it.


upendrakumbham (approved):
All looks good to me including the test. I understood, why we introduced boolean prop `raw` and accordingly changes to use fetchLoader to fetch text message instead of JSON.

PR #103, opened 2022-03-18
📦 scxa-marker-gene-heatmap 💬 Bugfix #181594058 – Fix cluster label not showing on the right

lingyun1010:
condition for plotting


packages/scxa-marker-gene-heatmap/src/MarkerGeneHeatmap.js
```diff
@@ -113,7 +109,7 @@ const MarkerGeneHeatmap = (props) => {
       }
     })
   })
-
+console.log(`plotLines`, plotLines)
```
alfonsomunozpomer, line 112: Please remove `console.log` statements.

alfonsomunozpomer (changes requested):
I tried the example pages with `npx webpack-dev-server -d` but they are broken. Could you please fix them?

alfonsomunozpomer:
@lingyun1010 Could you please reference the Pivotal Tracker ID in the title of the PR and describe what has been fixed, how and why? Otherwise this is very difficult to review. Thanks.

lingyun1010:
The pivotal tracker is https://www.pivotaltracker.com/story/show/181594058, we need to fix the label on the right is not showing properly which checking out a single cluster in the heatmap.

alfonsomunozpomer:
Please, have a look at the other two PRs if you need a template to apply to your PR title. The title of the PR is broken with your first comment:
image

Also, try to ellaborate on what the issue is, what was the underlying problem and how you solved it, to help code reviewers to folllow your thought process.

lingyun1010:
The label is not showing properly due to the mis-match between the heatmap data and the label list. I pass the filtered data as a prop and create a new list of label if the cluster ID/cell type is selected.

alfonsomunozpomer (commented):
I’m a bit confused about this PR. Under which conditions has `filteredData` the value of `null` so that the tertiary conditions in L65 and L75 will use `data` instead?

lingyun1010:
`data` will be used while the heatmap is not chunked by setting cluster ID/cell type in the dropdown box.

alfonsomunozpomer:
> `data` will be used while the heatmap is not chunked by setting cluster ID/cell type in the dropdown box.
 
That doesn’t answer my question. Put it another way: when is `filteredData` `null`?

lingyun1010:
When cluster is not ever selected, `filteredData` is `null`.

alfonsomunozpomer:
Look at this screenshot:
image

You can see in the bottom right pane that `selectedClusterId` is `null`, but `filteredData` is populated. So I don’t think that’s the case.

alfonsomunozpomer:
Just in case you were wondering, and to make things clearer, here’s `filteredData` in the heatmap component, rather than the view:
image

alfonsomunozpomer:
As far as I can see it changes when I click on the drop-down...

alfonsomunozpomer:
Now you have to ask yourself why do we want both `data` and `filteredData` if they had the same value. :wink:

lingyun1010:
They are not the same all the time. Once you select some cluster ID, `filteredData` is a trunk of `data`.

alfonsomunozpomer:
Yes, but do you really need both? The only thing you care for is to know if you are using `data` or `filteredData`, and the prop `isDataFiltered` tells you that, right?

alfonsomunozpomer:
So, before this PR, the component `ClustersHeatmapView` was setting the following prop for the heatmap:
```javascript
<MarkerGeneHeatmap
  data={filteredData}
  ...
/>
```

In your changes, you add the additional `data` array with unfiltered data and it is used for:
1. Compute `groupedData`
2. Use `groupedData` to get the `numberOfRows`
3. Use `numberOfRows` to calculate the Y offset of the label

So, ultimately, the only thing you want to know is how much offset you should apply to the label. The only piece of information you need to do that is to know if `isDataFiltered`, in which case you can just apply a fixed offset which will always be visible.

E.g.:
```javascript
const yOffset  = this.props.isDataFiltered ? -125 : -numberOfRows * rowHeight/2
```

Does that make sense?

lingyun1010:
> I'm assuming we are trying to resolve this issue ([ebi-gene-expression-group/atlas-web-single-cell#252](https://github.com/ebi-gene-expression-group/atlas-web-single-cell/issues/252)) here. If that's the case then I'm still seeing the issue. Can you please check?
> 
> Screenshot 2022-06-22 at 13 43 22

No. You may find the description below here. Basically, when user select cluster ID to display the label is not showing on the right except the first cluster.

> The pivotal tracker is https://www.pivotaltracker.com/story/show/181594058, we need to fix the label on the right is not showing properly which checking out a single cluster in the heatmap.

alfonsomunozpomer:
I can’t see the labels to the right in clusters:
image

They appear aggressively ellipsified in cell types (this is also mentioned in https://github.com/ebi-gene-expression-group/atlas-web-single-cell/issues/266, maybe it’s unrelated but it would be good maybe to fix this in this same PR?):
image

And they’re displayed broken in multiexperiment cell types:
image

Be more careful when merging code! Have a look at the latest PRs if in doubt. Also, ESLint the files you change so we end up having a homogeneous code style.

lingyun1010:
It shows properly on my local.
Screenshot 2022-07-11 at 10 20 25
Screenshot 2022-07-11 at 10 20 34

lingyun1010:
> I can’t see the labels to the right in clusters: image
> 
> They appear aggressively ellipsified in cell types (this is also mentioned in [ebi-gene-expression-group/atlas-web-single-cell#266](https://github.com/ebi-gene-expression-group/atlas-web-single-cell/issues/266), maybe it’s unrelated but it would be good maybe to fix this in this same PR?): image
> 
> And they’re displayed broken in multiexperiment cell types, [when they shouldn’t be shown at all](https://github.com/ebi-gene-expression-group/atlas-components/pull/115): image
> 
> Be more careful when merging code! Have a look at the latest PRs if in doubt. Also, ESLint the files you change so we end up having a homogeneous code style.

@alfonsomunozpomer I just reproduced your results by rebasing the master commits into my PR. Though I got the correct results in the PR branch, let me reimplement this based on the latest.

packages/scxa-marker-gene-heatmap/src/MarkerGeneHeatmap.js
```diff
@@ -176,13 +172,13 @@ const MarkerGeneHeatmap = (props) => {
       endOnTick: false,
       gridLineWidth: 0,
       minorGridLineWidth: 0,
-      plotLines: plotLines,
+      plotLines,
```
alfonsomunozpomer, line 175: :100:

packages/scxa-marker-gene-heatmap/src/MarkerGeneHeatmap.js
```diff
@@ -149,9 +145,9 @@ const MarkerGeneHeatmap = (props) => {
       opposite: true,
       categories: xAxisCategories,
       labels: {
-        useHTML: true,
+        useHtml: true,
```
alfonsomunozpomer, line 148: [The latest version of the Highcharts API has changed this attribute to be `useHTML`](https://api.highcharts.com/highcharts/xAxis.labels.useHTML).

packages/scxa-marker-gene-heatmap/src/MarkerGeneHeatmap.js
```diff
@@ -59,60 +59,56 @@ Highcharts.SVGRenderer.prototype.symbols.download = (x, y, w, h) => [
...
-      label: heatmapType !== `multiexperimentcelltypes` && {
-        useHTML: true,
-        text: splitCellTypeLabel.join(`<br/>`),
-        align: `right`,
-        textAlign: `left`,
-        x: 15,
-        y: yOffset - Math.max(splitCellTypeLabel.length - 3, 0) * rowHeight / 2 // Move very long labels slightly up
-      }
+  }) :
+    cellTypesOrClusterIds.map((cellTypeOrClusterId) => {
```
alfonsomunozpomer, line 91: `map` here isn’t returning any value, so either a different mathod or the function arugment should be changed.

packages/scxa-marker-gene-heatmap/src/MarkerGeneHeatmap.js
```diff
@@ -59,60 +59,56 @@ Highcharts.SVGRenderer.prototype.symbols.download = (x, y, w, h) => [
...
-    let color, zIndex
-    // don't show last plot line
-    if (idx === array.length - 1) {
-      // removing the plot line altogether would remove the label, so we need to make it "invisible"
-      color = `#FFFFFF`
-      zIndex = 0
-    } else {
-      color = `#000000`
-      zIndex = 5
+  cellTypesOrClusterIds.length === 1 ? plotLines.push({
```
alfonsomunozpomer, line 78: ESLint complains about this way of formatting the ternary operator.

packages/scxa-marker-gene-heatmap/src/MarkerGeneHeatmap.js
```diff
@@ -59,60 +59,56 @@ Highcharts.SVGRenderer.prototype.symbols.download = (x, y, w, h) => [
...
+      plotLines.push({
+        color,
+        width: 2,
+        value: plotLineAxisPosition,
+        zIndex,
+        label: heatmapType !== `multiexperimentcelltypes` && {
+          text: splitCellTypeLabel.join(`<br/>`),
+          align: `right`,
+          textAlign: `left`,
+          x: 0,
```
alfonsomunozpomer, line 107: Originally we added 15 pixels of separation between the heatmap and the labels. Could we have it back? You may need to change the options of the plot to reserve more space for the labels in that case.

alfonsomunozpomer (changes requested):
Some minor changes. Let’s get this set of improvements done and I’ll give it another go. Functionality is there, good job!

packages/scxa-marker-gene-heatmap/src/MarkerGeneHeatmap.js
lingyun1010: This is the bug fix to have the label back, so we can't reset it back to `15`.

alfonsomunozpomer: It turns out that there is a bug in Highcharts. I’ve reported it here:
https://github.com/highcharts/highcharts/issues/18149

In the meantime I’ve got a fix planned to add a bit of space to the label, but I’ll open a separate PR for that. :)

alfonsomunozpomer (approved):
At last! Merge away.

packages/scxa-marker-gene-heatmap/src/MarkerGeneHeatmap.js
alfonsomunozpomer: It was a duplicate of https://github.com/highcharts/highcharts/issues/17888 and there are two workarounds, which we’ll add once this is merged.

PR #105, opened 2022-06-01
:package: atlas-react-fetch-loader :speech_balloon: Feature #179948196 – Support query request parameters

alfonsomunozpomer:
Add a prop to support query parameters (e.g. `?foo=bar`) in requests. The prop is an object that [can be parsed by URI.js as-is](https://medialize.github.io/URI.js/docs.html#accessors-search).


lingyun1010 (approved):
Looks good! Very useful prop.

PR #106, opened 2022-06-07
:package: eslint-config :speech_balloon: Chore – Version package manually independently of Lerna

alfonsomunozpomer:
Unfortunately, Lerna’s way of symlinking packages makes our `eslint-config` not work properly. If you run `npx eslint src/*.js` you’ll get something like:
```
Oops! Something went wrong! :(

ESLint: 8.16.0

ESLint couldn't find the config "standard" to extend from. Please check that the name of the config is correct.

The config "standard" was referenced from the config file in "/home/amunoz/tmp/atlas-components/packages/eslint-config/index.js".

If you still have problems, please stop by https://eslint.org/chat/help to chat with the team.
```

I think it has to do to the way peer dependencies are handled by `npm`. This could be solved by manually installing `eslint-config-standard` in the package or as a global dependency in the project, but in the end it’s a kludge that needs to be maintained manually. If it requires manual maintenance it’s better to be explicit about it rather than play cat-and-mouse with Lerna.

Whatever the case, it didn’t seem sensible either to detect transitive changes because the ESLint config version changes, so there’s some added value.

I published the updated v3 and it can be used straight away.


haideriqbal (approved):
LGTM!

PR #107, opened 2022-06-07
:package:  scxa-cell-type-wheel :speech_balloon: Feature #179948196 – Add experiment accessions to click callback

alfonsomunozpomer:
Add the experiment accessions from the `data` array elements to the on click callback together with the cell type and the species.

Regarding maintenance, I cleaned up all the demo HTML pages, added real data, updated dependencies all around and put just the necessary code to at least test that the component renders properly. Unfortunately I couldn’t get much further because Highcharts is a big opaque blob of SVG-rendering native JS code which doesn’t let us test anything beyond that.


packages/scxa-cell-type-wheel/src/CellTypeWheel.js
```diff
@@ -97,23 +92,33 @@ class CellTypeWheel extends React.Component {
     }
   }
 
-  render() {
+  render () {
     const { chartOptions } = this.state
 
     return (
-      <div>
+      <figure>
```
haideriqbal, line 99: 💯

packages/scxa-cell-type-wheel/src/CellTypeWheel.js
```diff
@@ -97,23 +92,33 @@ class CellTypeWheel extends React.Component {
...
+        </figcaption>
+      </figure>
     )
   }
 }
 
 CellTypeWheel.propTypes = {
   searchTerm: PropTypes.string.isRequired,
-  data: PropTypes.object.isRequired,
+  data: PropTypes.arrayOf(
```
haideriqbal, line 114: 🎖️

haideriqbal (dismissed):
Code wise it all looks good! I'm trying to run this locally but getting the following error:

`Class constructor ServeCommand cannot be invoked without 'new'`

I believe this might be because of my local dependencies. I am looking into it and in the meanwhile, If you are able to run it on your end and tests are passing please merge it away.

haideriqbal:
After the update to lerna 5 I can run the component locally so all good!

PR #108, opened 2022-06-08
📦 scxa-tsne-plot 💬 Feature #181139004 – Move back 'Not available' metadata categories

lingyun1010:
Massive `Not available` points will block other categories like below:
Screenshot 2022-06-08 at 06 58 04

I added a z-index for each data series and make sure `Not available` is a negative and solve the issue:
Screenshot 2022-06-08 at 07 26 17

### DIFF
--- a/packages/scxa-tsne-plot/__test__/ClusterTSnePlot.test.js
+++ b/packages/scxa-tsne-plot/__test__/ClusterTSnePlot.test.js
@@ -53,7 +53,7 @@ describe(`ClusterTSnePlot colourize function`, () => {
     const randomSeries = randomHighchartsSeriesWithNamesAndMaxPoints([...seriesNames, `Not available`], maxPointsPerSeries)
     _colourizeClusters([], `lightgrey`)(randomSeries).forEach((series) => {
       if(series.name === `Not available`) {
-        expect(series).toHaveProperty(`zIndex`)
+        expect(series).toHaveProperty(`zIndex`, -1)
       } else {
         expect(series).not.toHaveProperty(`zIndex`)
       }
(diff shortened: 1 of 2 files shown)


alfonsomunozpomer (approved):
As usual, Highcharts documentation was a bit spotty. :smile:

alfonsomunozpomer (changes requested):
I just noticed that `_colourizeClusters` is being exported, and it’s tested in `ClusterTSnePlot.test.js`, so @lingyun1010 could you please add a test for this change?

packages/scxa-tsne-plot/__test__/ClusterTSnePlot.test.js
```diff
@@ -49,6 +49,17 @@ describe(`ClusterTSnePlot colourize function`, () => {
     })
   })
 
+  test(`must move behind if metadata is Not Available`, () => {
+    const randomSeries = randomHighchartsSeriesWithNamesAndMaxPoints([...seriesNames, `Not available`], maxPointsPerSeries)
+    _colourizeClusters([], `lightgrey`)(randomSeries).forEach((series) => {
+      if(series.name === `Not available`) {
+        expect(series).toHaveProperty(`zIndex`)
```
alfonsomunozpomer, line 56: Since we want the “Not available” series to be behind the other series, the test should check that the `zIndex` value is negative.

alfonsomunozpomer (commented):
My `npx jest` breaks after doing `npx lerna bootstrap`. @lingyun1010 Does this work for you?

alfonsomunozpomer (approved):
:+1:

PR #110, opened 2022-06-10
📦 eslint-config 💬 Add rules for tertiary operator

alfonsomunozpomer:
Just what the title says, some rules to enforce readable tertiaries.


haideriqbal (approved):
LGTM!

PR #111, opened 2022-06-11
📦 scxa-cell-type-wheel-experiment-heatmap 💬 Feature #179948196 – Filter heatmap by experiment accessions

alfonsomunozpomer:
**NOTE**: When reviewing this PR ignore all changes under the path `scxa-cell-type-wheel`. Those are part of https://github.com/ebi-gene-expression-group/atlas-components/pull/107 and should be reviewed independently.

Besides adding the `experimentAccessions` as request parameters to the fetch request, I followed the same strategy with both components (i.e. the wheel and the heatmap): enrich the wheel with our fetch loader HOC and avoid adding data as props in the wrapping component. This will require changes in the JSP but it will make the maintenance of this component a lot simpler, as changes in the wheel will require fewer changes here, if any.

Lastly I made the component fully functional, also following the latest React recommendations to use `function` rather than arrow functions, and declaring functions before `render` to avoid creating them every time the component re-renders.

I added one minimal test, not so much for correctness, but to easily append more tests.


lingyun1010:
I tried to run the demo, but once clicked a cell type, I got an error in the console `TypeError: f.default[l] is undefined`.

haideriqbal (changes requested):
I cannot see the marker gene heatmap after clicking the outer ring of the cell type wheel. Sharing the screenshots for details:

Screenshot 2022-06-13 at 15 38 28
Screenshot 2022-06-13 at 15 38 19
Screenshot 2022-06-13 at 15 38 10
Screenshot 2022-06-13 at 15 34 06

haideriqbal (approved):
LGMT! Console errors gone away and component is working fine. It was lerna bootstrap issue on my end.

PR #112, opened 2022-06-15
📦 scxa-marker-gene-heatmap 💬 Feature #181986522 - Add icon to ann data experiments

haideriqbal:
I had to add label formatting to the heatmap highchart config rather in `heatmapOptionsProvider` because the grouping labels are using a splitting label mechanism which treats the `html` as a string and as a result, the icon is not shown.

The downside of this approach is we cannot really test the added functionality. However, I have tested with dummy data on my local machine and it works fine.


packages/scxa-marker-gene-heatmap/src/MarkerGeneHeatmap.js
```diff
@@ -105,7 +109,11 @@ const MarkerGeneHeatmap = (props) => {
       value: plotLineAxisPosition,
       zIndex: zIndex,
       label: heatmapType !== `cellmultiexperiment` && {
-        text: splitCellTypeLabel.join(`<br/>`),
+        useHTML: true,
+        text:  checkIfAnnDataExperiment(splitCellTypeLabel) ?
+            `<img src="https://upload.wikimedia.org/wikipedia/commons/6/6a/External_link_font_awesome.svg" alt="external_link_image" width="11px"/>` +
+            splitCellTypeLabel.join(`<br/>`) :
+            splitCellTypeLabel.join(`<br/>`),
```
alfonsomunozpomer, line 116: This bit shouldn’t need changing since [we agreed to remove the experiment accessions from the right margin in the multiexperiment cell type heatmap](https://www.pivotaltracker.com/story/show/179903243). We had a regression there and it has been fixed in https://github.com/ebi-gene-expression-group/atlas-components/pull/115.

packages/scxa-marker-gene-heatmap/src/clustersView/ClustersHeatmapView.js
```diff
@@ -77,7 +77,7 @@ class HeatmapView extends React.Component {
   render() {
     const { data, filteredData, isLoading, hasError } = this.state
     const { wrapperClassName, plotWrapperClassName } = this.props
-    const { ks, ksWithMarkers, selectedClusterId, onChangeMarkerGeneFor, onChangeClusterId, metadata, cellTypes, selectedClusterCategory, selectedClusterId} = this.props
+    const { ks, ksWithMarkers, onChangeMarkerGeneFor, onChangeClusterId, metadata, cellTypes, selectedClusterCategory, selectedClusterId} = this.props
```
alfonsomunozpomer, line 80: :detective:

packages/scxa-marker-gene-heatmap/src/MarkerGeneHeatmap.js
```diff
@@ -149,9 +157,12 @@ const MarkerGeneHeatmap = (props) => {
...
       categories: xAxisCategories,
       labels: {
-        useHtml: true,
+        useHTML: true,
         formatter: function() {
-          return heatmapOptionsProvider[heatmapType].labelsFormatter(this.value)
+          return checkIfAnnDataExperiment(this.value) ?
+              `<img src="https://upload.wikimedia.org/wikipedia/commons/6/6a/External_link_font_awesome.svg" alt="external_link_image" width="11px"/> ` +
+              heatmapOptionsProvider[heatmapType].labelsFormatter(this.value) :
+              heatmapOptionsProvider[heatmapType].labelsFormatter(this.value)
```
alfonsomunozpomer, line 165: This logic should be moved to `heatmapOptionsProvider.js` as we discussed.

alfonsomunozpomer (changes requested):
Besides the two changes mentioned, could you please remember to follow the template for the PRs?

📦 <package-name> 💬 {Feature|Chore|Bugfix} #<ID> – Description

Also, try to remember to prefix the branch name with the package you are working on (this is less important but it’s a good practice).

alfonsomunozpomer (approved):
All good! Next stop will be at the Add Tests station. But that will be a battle fought another day.

alfonsomunozpomer:
I merged `master` into this feature branch, adjusting the code to render both the SVG icon and the link to the experiment page. I added an ANND experiment accession to the demo HTML pages, so at least we can check the result visually.

As sidenote, remember that:
- We don’t use semicolons (now with ESLint 8 and our reviewed config you’ll be reminded if you have them)
- Arrow functions with only one statement don’t need to `return` the result, it’s implicitly done
- [Regexes should be created with slashes rather than with the `new` constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#creating_a_regular_expression) unless, of course, it’s dynamic

alfonsomunozpomer (approved):

PR #115, opened 2022-06-27
📦 scxa-marker-gene-heatmap 💬 Feature #179903224, #179903243 – Add link to experiment accession labels and fix regression

alfonsomunozpomer:
Besides linking the column label I noticed we inadvertently had put back the experiment accessions to the right of the multiexperiment cell type heatmap, when we shouldn’t have. The check `heatmapType !== \`cellmultiexperiment\`` was of course failing to match anything. Maybe we should keep track of this sort of hard-coded strings some other way, maybe using `Object.keys`, or some other way to avoid this type of mistakes (I suspect I changed the name in the provider and forgot to change it elsewhere).


alfonsomunozpomer:
Besides linking the column label I noticed we inadvertently had put back the experiment accessions to the right of the multiexperiment cell type heatmap, when we shouldn’t have. The check `heatmapType !== \`cellmultiexperiment\`` was of course failing to match anything. Maybe we should keep track of this sort of hard-coded strings some other way, maybe using `Object.keys`, or some other way to avoid this type of mistakes (I suspect I changed the name in the provider and forgot to change it elsewhere).

PR #117, opened 2022-10-14
Add scxa-marker-gene-widget component

ke4:


ke4:
We don't need this new component as we already have it.

PR #118, opened 2022-10-19
📦 scxa-marker-gene-heatmap 💬 Add tests with Cypress

alfonsomunozpomer:
Highcharts breaks our tests: anything that goes beyond just checking rendering with static props throws a bunch of exceptions due to the way Highcharts handles updates. They are caused by some mis-handling of refs by js-dom. In fact, [even the tests included in `highcharts-react-official` are failing](https://github.com/highcharts/highcharts-react/issues/384).

After extensive trial and error Cypress is the best thing we’ve found. So here it is! I’ve added [code coverage via Istanbul following the official Cyrpess guide](https://docs.cypress.io/guides/tooling/code-coverage).


ke4 (approved):
LGTM 💯

packages/scxa-cell-type-wheel/README.md
```diff
@@ -1,84 +1,14 @@
...
-## Polyfills
-No polyfills are included by default, but you might want one or both of these:
-- [Fetch polyfill](https://github.com/github/fetch)
-- [Babel polyfill](https://babeljs.io/docs/usage/polyfill/)
-
-### NPM
+If you want to omit video recording:
 ```bash
-npm install --save-dev whatwg-fetch @babel/polyfill
+npx cypress run --component
```
ke4, line 10: This is the same command as above. Maybe you forgot to add a switch/parameter here?

packages/scxa-cell-type-wheel/README.md
```diff
@@ -1,84 +1,14 @@
...
-  ...
-}
-```
-
-## Run it on your browser
-Use Webpack-Dev-Server:
-```bash
-npx webpack-dev-server -d
-```
+Find coverage reports at:
```
ke4, line 13: I think we should add the README on how to achieve coverage reporting. What are the steps to do for this?

ke4 (changes requested):
Tiny change requests

packages/scxa-cell-type-wheel/README.md
alfonsomunozpomer: Oh, right!

packages/scxa-cell-type-wheel/README.md
alfonsomunozpomer: Fixed.

ke4 (approved):
Perfect ;-)  👍

PR #119, opened 2022-10-19
📦 eslint-config 💬 Adjust rules for spacing argument block in fucntion declarations

alfonsomunozpomer:
Add a rule that follows our coding style. You can see some examples at https://eslint.org/docs/latest/rules/space-before-function-paren#anonymous-always-named-never-asyncarrow-always.


ke4 (approved):
LGTM

PR #122, opened 2022-11-01
[Manual version update] Cell type wheel experiment heatmap - add search form

ke4:
Reason for the manual update:
I already pushed these package changes to npm manually as I had an npm login problem when I tried to use `npx lerna publish` for the first time. I had permission issues that we fixed later.

Changed packages:
- @ebi-gene-expression-group/scxa-cell-type-wheel-heatmap
- @ebi-gene-expression-group/scxa-cell-type-wheel
- @ebi-gene-expression-group/scxa-marker-gene-heatmap


ke4:
We don't need this PR. We fixed this issue in another way with @alfonsomunozpomer .
Thanks for your help @alfonsomunozpomer 😉

PR #123, opened 2022-11-01
Fix search field height in the demo

ke4:
Changes:

- Fixed the CSS properties for the search text field in the demo for the cell type wheel heatmap component.
- Updated for the latest eslint config (`3.5.1`)


alfonsomunozpomer (approved):
LGTM.

PR #124, opened 2022-11-10
 📦 scxa-faceted-search-results 💬 Replace broken enzyme tests with Cypress ones

ke4:
Changes: 
- Replace broken enzyme unit tests with Cypress ones
- Upgrade versions of dependencies
- Added code coverage with Cypress


lingyun1010:
In the meanwhile, please add a corresponding description in the README file, including the npm version (should be up from v7, otherwise there will be a configuration error), and the command line to start a cypress test window.

ke4:
@lingyun1010 Thanks a lot for your suggestions. I updated the `README` with them. 
Please take a look if it is any better. Thanks 👍

packages/scxa-faceted-search-results/cypress/component/FacetedSearchContainer.cy.js
```diff
@@ -0,0 +1,105 @@
...
+          uniqueness = false
+        }
+      }).then(() => {
+        expect(uniqueness).is.true
+      })
+
+    let dropdownTitles = []
+    cy.get(`div.css-qbdosj-Input`).first().click()
+    cy.get(`div.css-dzz542-menu`).then(() => {
+      cy.get(`div.css-1n6sfyn-MenuList`).then((selectedTitles) => {
```
alfonsomunozpomer, line 85: I strongly encourage to use a different set of selectors here.

This is not very robust because the CSS classes’ IDs are generated by the styling framework used by React Select. If the framework decides that these IDs should be generated in a different way, or we choose to style the component differently, the test will probably break.

Another issue is that a CSS class is being used to select an element but no assertion about its style is made, so it doesn’t seem that the intent is clear.

For example, you can change L83 to:
```javascript
cy.get(`input#facetGroupMultiSelectDropdown`).first().click()
```

ke4: That would be my exact comment, too. I don't want to use a randomly generated selector. That could break later.

alfonsomunozpomer:
[How about adding code coverage?](https://github.com/ebi-gene-expression-group/atlas-components/tree/master/packages/scxa-cell-type-wheel#add-code-coverage-in-a-nutshell)

packages/scxa-faceted-search-results/cypress/component/FacetedSearchContainer.cy.js
alfonsomunozpomer: Also, if there’s no suitable alternative, [you can use regexes in CSS selectors](https://stackoverflow.com/questions/8903313/using-regular-expression-in-css). I’ve noticed that the classes have suffixes such as `-menu` or `-MenuList`.

alfonsomunozpomer (changes requested):
[How about adding code coverage?](https://github.com/ebi-gene-expression-group/atlas-components/tree/master/packages/scxa-cell-type-wheel#add-code-coverage-in-a-nutshell)

upendrakumbham (approved):
LGTM!

PR #133, opened 2023-01-25
Fix cell type data label not showing & add a test

lingyun1010:
Cell type data labels are hidden while in a small window for example Homo Sapiens > brain. I fixed it by disabling the `rotationMode` and add a cypress test for it.

### DIFF
--- a/packages/scxa-cell-type-wheel/cypress/component/CellTypeWheel.cy.js
+++ b/packages/scxa-cell-type-wheel/cypress/component/CellTypeWheel.cy.js
@@ -2,7 +2,9 @@ import React from 'react'
 import CellTypeWheel from '../../src/CellTypeWheel'
 
 import cellTypeWheelData from '../fixtures/lung-cancer.json'
-import bigCellTypeWheelData from '../fixtures/cancer.json'
+// cell type wheel data labels not showing properly with the cancer json response on Mac
+// a github issue is reported in highcharts repo: https://github.com/highcharts/highcharts/issues/18332
+import cancerCellTypeWheelData from '../fixtures/cancer.json'
 
 // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random
 function getRandomInt(max) {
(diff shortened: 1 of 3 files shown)


packages/scxa-cell-type-wheel/cypress/component/CellTypeWheel.cy.js
```diff
@@ -2,6 +2,7 @@ import React from 'react'
 import CellTypeWheel from '../../src/CellTypeWheel'
 
 import cellTypeWheelData from '../fixtures/lung-cancer.json'
+import bigCellTypeWheelData from '../fixtures/cancer.json'
```
alfonsomunozpomer, line 5: Now that we have two payloads, we can give them proper names such as `lungCancerCellTypeWheelData` or `cancerCellTypeWheelData`. It may be useful to comment in the test, or here in the `import`, the specific browser and OS where this unusual situation happens, and a link to the GitHub issue in Highcharts.

packages/scxa-cell-type-wheel/cypress/component/CellTypeWheel.cy.js
```diff
@@ -89,4 +90,26 @@ describe(`<CellTypeWheel>`, () => {
 
     cy.get(`@onCellTypeWheelClick`).should(`not.have.been.called`)
   })
+
+  it(`should show data labels properly while zooming in from an inner cell`, () => {
```
alfonsomunozpomer, line 94: I think this describes the test more accurately: “should show data labels properly after clicking on a species/organism part and zooming in”.

packages/scxa-cell-type-wheel/cypress/component/CellTypeWheel.cy.js
```diff
@@ -89,4 +90,26 @@ describe(`<CellTypeWheel>`, () => {
...
+    const callbackWrapper = {
+      onCellTypeWheelClick: function (cellType, species, experimentAccessions) {
+      }
+    }
+    cy.mount(
+        <CellTypeWheel
+            data={bigCellTypeWheelData}
+            onCellTypeWheelClick={callbackWrapper.onCellTypeWheelClick}
+        />)
+    let innerRingElement = cy.contains(`Homo sapiens`)
```
alfonsomunozpomer, line 104: This can be `const`. As a rule of thumb use `const`, unless you can’t.

alfonsomunozpomer (changes requested):
Minor details. Address them and this is good to go. :+1

ke4 (dismissed):
LGTM 👍

alfonsomunozpomer (approved):
All good!

PR #135, opened 2023-02-03
Add a visual UI distinction for anndata experiments in the experiments table

lingyun1010:
We add an icon column for experiment table to mark if it is anndata or not. We don't have a proper endpoint for anndata experiment type yet, so I pick the frontend solution. A jest test is added for the new feature.

### DIFF
--- a/packages/atlas-experiment-table/src/TableCell.js
+++ b/packages/atlas-experiment-table/src/TableCell.js
@@ -11,6 +11,10 @@ const TableCell = ({ dataRow, dataKey, image, linkTo, host, width }) => {
     if (image[dataRow[dataKey]]) {
       cellItem = <img {...image[dataRow[dataKey]]}/>
     }
+    else if(Object.keys(image).includes(`anndata`)){
+      let ifAnndata = dataRow[dataKey].substring(0,6) === `E-ANND` ? `anndata` : `nonAnndata`
+      cellItem= <img {...image[ifAnndata]}/>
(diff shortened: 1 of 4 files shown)


packages/atlas-experiment-table/src/TableCell.js
```diff
@@ -11,6 +11,10 @@ const TableCell = ({ dataRow, dataKey, image, linkTo, host, width }) => {
     if (image[dataRow[dataKey]]) {
       cellItem = <img {...image[dataRow[dataKey]]}/>
     }
+    else if (Object.keys(image).includes(`anndata`)) {
+      let ifAnndata = dataRow[dataKey].substring(0,6) === `E-ANND` ? `anndata` : `nonAnndata`
```
haideriqbal, line 15: I believe this check can be improved using a regular expression as compared to the hard-coding substring index. I guess you just want to see if the experiment accession has `E-ANND` in it? if yes, have a look [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test) to see if this check can be implemented using regex.

haideriqbal (dismissed):
Just one comment regarding the experiment check. Please follow our standard formatting template for atlas components. For example in this [PR.](https://github.com/ebi-gene-expression-group/atlas-components/pull/134)

upendrakumbham (approved):
ALGTM!

PR #136, opened 2023-02-03
Add a visual UI distinction for anndata experiments in the gene search results

lingyun1010:
Gene search result card added a new column to experiment type icon to distinguish anndata and the others. A tooltip is added if users hover on the icon. This is not a src code implementation, so there is no new test for it. Our component is compatible with new header and card style.


packages/scxa-faceted-search-results/html/fetch-loader/ExperimentCard.js
```diff
@@ -58,7 +66,7 @@ class ExperimentCard extends React.Component {
   }
 
   render() {
-    const {url, species, experimentDescription, markerGenes, numberOfAssays, factors} = this.props
+    const {url, species, experimentDescription, markerGenes, numberOfAssays, experimentAccession, type, factors} = this.props
```
ke4, line 69: I might miss something, but I can't see that the `experimentAccession` and `type` is being added to the `props`. Am I wrong?

ke4 (changes requested):
Hi, I added 2 comments. Could you review them, please?
Thanks

ke4 (approved):
Thanks for the changes

PR #140, opened 2023-06-08
Fix tsne and umap format and default parameter

lingyun1010:
We need to fix two things here:
1. `tsne` and `umap` from cell plot endpoint changed to `t-SNE` and `UMAP`
2. the default parameter and method props from experiment page endpoint is `defaultPlotMethodAndParameterisation` 
TODO: 3. Make the dropdown options dynamiclly rather than hard coded values for `t-SNE` or `UMAP`

Reviewer may test this PR under the latest postgres database.

### DIFF
--- a/packages/scxa-tsne-plot/html/Demo.js
+++ b/packages/scxa-tsne-plot/html/Demo.js
@@ -78,9 +78,8 @@ const experiment5 = {
 // Number of cells: 101,843
 const experiment6 = {
   accession: `E-MTAB-5061`,
-  accessKey: `f60a21b8-990a-49d9-95aa-623c10865faa`,
   species: `Homo sapiens`,
-  ks: [20],
+  ks: [8,11,19,21],
   metadata: [
     {
       value: `inferred cell type - authors labels`,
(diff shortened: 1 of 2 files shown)


upendrakumbham (dismissed):
ALGTM!

upendrakumbham:
@lingyun1010, I managed to install nvm and switch to node version 14. This resolved my front-end issue as you said before.

Here is the screenshot:

image

Alfonso help me in further debugging this issue and picking the right commit of atlas-web-core.

These endpoints still using old plot methods as data, and our demo page still contains old data. Could you please fix this data as per the latest plot method changes? Let me know once you finished.

Thanks

upendrakumbham:
Hi @lingyun1010, you're right I was using master branch for the testing. I switched to above branch and tested. All working fine after a small change in the `demo.js `. I have accesskey in my DB for the `E-MTAB-5061` is,  this one `7ff78b93-f081-4087-9bd3-abcff0ef0b60`, but demo page has different one. Not sure which is correct one. Please verify from your end before merging, other wise all good. So I'm approving both PR's.

upendrakumbham (dismissed):
All looks good. Tested PR with the corresponding backend on my local machine. Working as expected.

upendrakumbham (approved):
All looks good!

PR #141, opened 2023-06-22
atlas-experiment-table – Enable the component to make requests for the footer links

lingyun1010:
This story refers to the frontend implementation of the new experiment design logic. We need to change the way React router is populating the experiment page. Currently, all of the payload is delivered in one go but we will need to change that behavior and utilize new endpoints implemented on the backend for experiment design.

The expected behaviour is when a user clicks on the Experiment Design tab, a request is made to backend the experiment design controller and the experiment design table is populated. New controllers return the payload for experiment design in the same format as previously so we might not need to change how the data is parsed in the frontend component but we do need to handle edge cases like maintaining pagination functionality, and finding when the end of file has been reached.

Relevant back end PRs for controller information are [scxa](https://github.com/ebi-gene-expression-group/atlas-web-single-cell/pull/324), [gxa](https://github.com/ebi-gene-expression-group/atlas-web-bulk/pull/151) and [web-core](https://github.com/ebi-gene-expression-group/atlas-web-core/pull/123).


upendrakumbham (approved):
All looks good to me!

PR #142, opened 2023-08-07
Atlas experiment table/adjust number of table headers and header height dynamically

lingyun1010:
As we have a secondary table header in experiment design table, some experiment may contains a long title which requires an adjustable header height


ke4 (approved):
LGTM

PR #144, opened 2023-10-25
Bugfix/colour by category url logic 400 error

lingyun1010:
I fix the logic in the url fetching stage, which is more straight forward, as the `colourByCategory` is decided by the vlaue of `colourBy`

### DIFF
--- a/packages/scxa-tsne-plot/src/TSnePlotView.js
+++ b/packages/scxa-tsne-plot/src/TSnePlotView.js
@@ -57,7 +57,7 @@ class TSnePlotView extends React.Component {
 
   _fetchAndSetStateCellClusters(
     {atlasUrl, experimentAccession, accessKey, selectedColourBy, selectedColourByCategory, selectedPlotType, selectedPlotOptionLabel}) {
-    const resource = selectedColourByCategory === `clusters` ?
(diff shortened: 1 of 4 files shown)


packages/scxa-tsne-plot/src/TSnePlotView.js
```diff
@@ -57,7 +57,7 @@ class TSnePlotView extends React.Component {
 
   _fetchAndSetStateCellClusters(
     {atlasUrl, experimentAccession, accessKey, selectedColourBy, selectedColourByCategory, selectedPlotType, selectedPlotOptionLabel}) {
-    const resource = selectedColourByCategory === `clusters` ?
+    const resource = !isNaN(selectedColourBy) ?
```
ke4, line 60: Is there a test for this code change? Thanks

ke4 (changes requested):
I am missing a test for the code change.
Could you please add it or please describe me why we don't need it?
Thanks

upendrakumbham (approved):
ALGTM!

lingyun1010:
> I am missing a test for the code change. Could you please add it or please describe me why we don't need it? Thanks

I have looked into this issue. The true or false logic just change the value of a variable not affect any other props or state or UI staff, so we don't have a test for this.

PR #147, opened 2024-05-08
Atlas experiment table/add anndata to exp title column

lingyun1010:
We changed the requirement to add an Anndata icon to the experiment table title column in SC, rather than adding an image column like what we have in bulk experiment page. If the experiment is not an `E-ANND` type, then no icons should be shown as discussed with @Silvie.

We need to make the experiment table component compilable with bulk and SC pages.

Anndata icon image should be added into the backend `resource` as `differential` icons in bulk. For now, I am just using the source link `https://www.ebi.ac.uk/gxa/resources/images/experiments-table/differential.png` for anndata at this development stage. We will need to update this in the javascript bundle in `atlas-web-single-cell` repository to integrit this change to SC application.

FYI: A test Anndata experiment is in page 13 as shown in the screenshot below.
Screenshot 2024-05-08 at 17 01 11

### DIFF
--- a/packages/atlas-experiment-table/__test__/TableCell.test.js
+++ b/packages/atlas-experiment-table/__test__/TableCell.test.js
@@ -65,28 +65,29 @@ describe(`TableCell`, () => {
 
   test.each(
       [`E-ANND-12`, `testExperimentAccession`, `E-MTAB-12`]
-  )(`is an image and check if it is an Anndata experiment or else`, (arrayDataKey) => {
-      const image = {
-            anndata: {
-              // tbc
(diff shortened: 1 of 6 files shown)


ke4 (approved):
LGTM!
Thanks for your work!
After the Team meeting, we should create a tiny PR to replace the icon with the agreed one.

lingyun1010:
According to the discussion from team meeting on 20th May, we changed the design for anndata experiment, with `A` css icon behind the experiment title and a detailed tooltip.
Screenshot 2024-05-20 at 17 08 16

PR #148, opened 2024-05-13
Move anndata icon to title column in gene search result page

lingyun1010:
As discussed with @silvie, we changed the UI design for anndata experiment in the sc gene search results page. We used to have a column to show anndata icons, now we add an small icon in the title column as shown in the screenshot below. The detailed css may need to be further discussed with other team members.

anndata-gene-search-result


ke4 (changes requested):
LGTM, but we need to replace the anndata icon to the one we discussed on this week's meeting. Other than that it is fine.
Thanks

lingyun1010:
According to the discussion from team meeting on 20th May, we changed the design for anndata experiment, with A css icon behind the experiment title and a detailed tooltip.
Screenshot 2024-05-20 at 17 47 10

ke4 (approved):
LGTM!
Thanks for your cleanup! It is much better!

PR #150, opened 2024-05-24
Atlas experiment table/feature/simplify the anndata icon loading without backend load change

lingyun1010:
While I was trying to integrate the changes to the backend, I found that in the current implementation for anndata, we require a new prop in the title data which is passed from backend, but it is not necessary and redundant, so I change the logic for anndata experiment icon. 

In a word, rather declaring `icon` from the backend, we just found anndata experiment and add an icon to `Title` column. The UI remains the same anyway.


ke4 (approved):
LGTM!
Thanks a lot for your work!

ke4 (commented):
LGTM!
Great work!

PR #154, opened 2024-07-23
Species search cell type wheel not clickable

lingyun1010:
In `scxa-cell-type-wheel` component, we hard-coded to make a gene heatmap search at node level 4, that's right for cell type metadata search, rather than for `species` metadata search. We need to make a condition to check if the search term is `species` or not by fetching all possible species from the `scxa-cell-type-wheel-experiment-heatmap` component.

If you want to test it locally, please remember to build a local `scxa-cell-type-wheel` package by typing `npm pack` under its directory and install with the absolute/relative directory in `scxa-cell-type-wheel-experiment-heatmap` component.

I also created a html demo page for `Mus muscuslus` search term in `scxa-cell-type-wheel` component, with altered json response, by removing the incorrect entry with `Homo sapiens`. This entry issue will be solved by data side as far as I know.

### DIFF
--- a/packages/scxa-cell-type-wheel-experiment-heatmap/src/CellTypeWheelExperimentHeatmap.js
+++ b/packages/scxa-cell-type-wheel-experiment-heatmap/src/CellTypeWheelExperimentHeatmap.js
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { useEffect, useState } from 'react'
 import PropTypes from 'prop-types'
 import URI from 'urijs'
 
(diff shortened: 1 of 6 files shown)


upendrakumbham (approved):
LGTM! I tested in my local, 3rd level(cell types) in the cell type wheel is clickable.

packages/scxa-cell-type-wheel-experiment-heatmap/src/CellTypeWheelExperimentHeatmap.js
```diff
@@ -14,17 +14,45 @@ const MarkerGeneHeatmapFetchLoader = withFetchLoader(MarkerGeneHeatmap)
...
+  const [error, setError] = useState(null)
+
+  useEffect(() => {
+    // Define the async function inside useEffect
+    const fetchData = async () => {
+      const url = URI(props.searchFormResource, props.host).toString()
+      try {
+        const response = await fetch(url)
+        if (!response.ok) {
+          throw new Error('Network response was not ok')
```
ke4, line 28:
```suggestion
          throw new Error('Network response was not ok. Please try again later.')
```

ke4 (changes requested):
I added a small suggestion, but otherwise it looks fine with me.

ke4 (approved):
LGTM

PR #155, opened 2024-07-25
[bugfix] Render big dataset with Highcharts

ke4:
The version of Highcharts 3rd party UI component that we are using to display charts had a problem in the earlier versions when it had to render more than 1000 data points.
I tested it with the latest version and it works fine. It could render the cell type wheel for the `Homo sapiens` search term. Earlier version only printed a confusing message to the display: `No data to display`.
I also added a test with the current `Homo sapiens` result set.


lingyun1010 (approved):
LGTM！

PR #158, opened 2024-08-05
Bugfix/species filter is getting ignored in the metadata search

upendrakumbham:
Bug: Our cell type wheel is not refreshing with the latest species drop-down value change. Simply, the wheel is ignoring species dropdown.

This PR -  Fixes that issue

To Fix this:

I have modified two components:

1. scxa-gene search Form 
2. scxa-scxa-cell-type-wheel-experiment-heatmap

Changes:

- Added a `onchange` function to notify the gene search form when species changed from the species drop-down
- Set  the latest species value in the state
- Trigger cell type wheel end point with latest species (Refresh cell type wheel component with latest data when we change species)
- passed species prop to the scxa-cell-type-wheel-experiment-heatmap component in the demo.js to refresh wheel every on change trigger

Local testing:
 - `'npm pack'` in the scxa-gene search Form  component and` tgz` file will be generated in that component after npm pack
-  copy the absolute path and do `npm  install <path>` in the scxa-cell-type-wheel-experiment-heatmap component
- 
- run `npx webpack-dev-server --mode=development`

### DIFF
--- a/packages/scxa-cell-type-wheel-experiment-heatmap/src/CellTypeWheelExperimentHeatmap.js
+++ b/packages/scxa-cell-type-wheel-experiment-heatmap/src/CellTypeWheelExperimentHeatmap.js
@@ -89,7 +89,7 @@ function CellTypeWheelExperimentHeatmap(props) {
           {props.searchTerm.trim() ?
             <CellTypeWheelFetchLoader
               host={props.host}
-resource={URI(props.cellTypeWheelResource)
+              resource={URI(props.cellTypeWheelResource)
                 .segment(props.searchTerm)
                 .toString()}
               query={props.species ? `?species=` + selectedSpecies : ''}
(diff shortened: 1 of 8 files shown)


upendrakumbham:
FIxed both, thanks for the comments.

packages/scxa-cell-type-wheel-experiment-heatmap/src/CellTypeWheelExperimentHeatmap.js
```diff
@@ -77,17 +82,17 @@ function CellTypeWheelExperimentHeatmap(props) {
...
         defaultSpecies={props.species}
+        onSpeciesSelectOnChange={handleChange}
       />
       <div className={`row-expanded small-12 columns`}>
         <div className={`small-12 medium-6 columns`} aria-label={`Cell type wheel`}>
           {props.searchTerm.trim() ?
             <CellTypeWheelFetchLoader
               host={props.host}
-              resource={URI(props.cellTypeWheelResource)
+resource={URI(props.cellTypeWheelResource)
```
ke4, line 92: I know it is a tiny formatting thing, but could you restore the aligning of this line, please?

ke4 (changes requested):
I am picky here and sorry for that, but could you restore a little bit formatting that I commented, please?
Otherwise all good. Thanks for your work!

packages/scxa-cell-type-wheel-experiment-heatmap/src/CellTypeWheelExperimentHeatmap.js
upendrakumbham: sorry, not sure, how it was disturbed, it makes sense. Thanks for that.

upendrakumbham: I fixed formatting, please feel free to review when log in

ke4 (approved):
LGTM, sorry about my pickiness

PR #160, opened 2024-08-07
Bugfix/species filter is getting ignored in the metadata search

upendrakumbham:
In this PR: I have updated only the query prop statement.

We need to review only the [last commit ](383b988f619a2606d0ea51ca13d9be50ff32b783)


ke4 (approved):
LGTM!

PR #162, opened 2024-08-09
[Chore] Get expression unit for TSNE plot from our db

ke4:
When I started to work on this fix the tests for `scxa-tsne-plot` components were broken.
I fixed the `snapshot` and added a missing dependency.

Some tests are still broken. WIP


lingyun1010:
Hi @ke4 thanks for your work, I just figured out why some tests are broken. It is because we add `exporting` wrapper for `highcharts` but it is not supported in the test. There is also a comment in `ScatterPlot.js` as 
```  
// Only include modules if Highcharts isn’t a *good* mock -- Boost/Exporting can break tests
```
, so if we would like to test, we may need to uncomment these lines first,
```
  HighchartsExporting(Highcharts)
  HighchartsOfflineExporting(Highcharts)
  ```

lingyun1010 (approved):
The tests are passed with updated snapshots and with comment out wrappers for highcharts. Besides, I publish new versions by lerna, as lerna has detected many changed files from history.

PR #167, opened 2024-08-21
Add cell type marker gene heatmap with dropdown menus

lingyun1010:
Please refer to the issue here https://github.com/ebi-gene-expression-group/atlas-web-single-cell/issues/441

### DIFF
--- a/packages/scxa-marker-gene-heatmap/src/clustersView/ClustersHeatmapView.js
+++ b/packages/scxa-marker-gene-heatmap/src/clustersView/ClustersHeatmapView.js
@@ -79,8 +79,8 @@ class HeatmapView extends React.Component {
   render() {
     const { data, filteredData, selectedClusterId, isLoading, hasError } = this.state
     const { wrapperClassName, plotWrapperClassName } = this.props
-    const { ks, ksWithMarkers, selectedK, onSelectK } = this.props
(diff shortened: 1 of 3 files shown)


packages/scxa-marker-gene-heatmap/src/clustersView/ClustersHeatmapView.js
```diff
@@ -90,16 +90,43 @@ class HeatmapView extends React.Component {
...
+    const options = [{
+      label: `Number of clusters`,
+      options: kOptions
+    }]
+
+    if (metadata) {
+      const metadataOptions = metadata.map((metadata) => ({
+        ...metadata,
+        // only enable inferred cell type metadata option and disable the others like sex, cancer etc.
+        isDisabled: ![`Inferred cell type - authors labels`, `Inferred cell type - ontology labels`].includes(metadata.label),
```
ke4, line 102: I would move these options to a variable and give them a descriptive name, then you could remove the comments in the previous line. The code would be cleaner with that change and more understandable for a new starter or even for us later when we are reading this code.
Thanks

ke4 (changes requested):
Just a tiny code cleaning, but otherwise looks fine.
Thanks for your work!

ke4 (approved):
Thanks for the changes, LGTM!

PR #170, opened 2024-09-09
Add no data message for cell type wheel

lingyun1010:


packages/scxa-cell-type-wheel/src/CellTypeWheel.js
```diff
@@ -4,10 +4,16 @@ import PropTypes from 'prop-types'
...
 
 import _ from 'lodash'
 
-highchartsSunburst(Highcharts)
+async function addModules () {
+  HighchartsNoData(Highcharts)
+  highchartsSunburst(Highcharts)
+}
+
+addModules()
```
upendrakumbham, line 16: Do you need this `addModules()` function call here?

packages/scxa-cell-type-wheel/src/CellTypeWheel.js
```diff
@@ -46,6 +52,17 @@ class CellTypeWheel extends React.Component {
           }
         },
 
+        lang: {
+          noData: `${props.species} doesn't have wheel data for the organism part: ${props.searchTerm}.`
```
ke4, line 56:
```suggestion
          noData: `No data exists in our database for this search term: ${props.searchTerm} for ${props.species}.`
```

ke4 (approved):
I just suggested to change the `noData` message, but otherwise looks fine by me

upendrakumbham (commented):
only tiny clarification on this `addModule()`, otherwise all looks good to me to approve.

packages/scxa-cell-type-wheel/src/CellTypeWheel.js
lingyun1010: I don't know, I think that was agreed in the team meeting, so @upendrakumbham wrote it down in the issue. https://github.com/ebi-gene-expression-group/atlas-web-single-cell/issues/455#issue-2478240961

PR #174, opened 2024-09-20
Add host to the link function for multi experiment

lingyun1010:
Easy fix, host prop is missing

### DIFF
--- a/packages/scxa-marker-gene-heatmap/__test__/heatmapOptionsProvider.test.js
+++ b/packages/scxa-marker-gene-heatmap/__test__/heatmapOptionsProvider.test.js
@@ -1,97 +1,96 @@
-import URI from 'urijs'
 import heatmapOptionsProvider, { checkIfAnnDataExperiment } from '../src/heatmapOptionsProvider'
 
 describe(`checkIfAnnDataExperiment`, () => {
-    it(`should return true for valid AnnData experiment accession`, () => {
-        expect(checkIfAnnDataExperiment(`E-ANND-1234`)).toBe(true)
-    })
+  it(`should return true for valid AnnData experiment accession`, () => {
+    expect(checkIfAnnDataExperiment(`E-ANND-1234`)).toBe(true)
+  })
 
-    it(`should return false for invalid AnnData experiment accession`, () => {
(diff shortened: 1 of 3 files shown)


ke4 (changes requested):
You modified the code, but have not modified any tests.
That could mean 2 things: either at least 1 test is broken or there is no tests for that function you modified.
Could you investigate and fix this, please?
Thanks

packages/scxa-marker-gene-heatmap/__test__/heatmapOptionsProvider.test.js
```diff
@@ -0,0 +1,96 @@
...
+
+describe(`heatmapOptionsProvider.multiexperimentcelltypes.labelsFormatter`, () => {
+  const labelsFormatter = heatmapOptionsProvider.multiexperimentcelltypes.labelsFormatter
+
+  it(`should return formatted label with AnnData icon for AnnData experiments`, () => {
+    const experimentAccession = `E-ANND-0001`
+    const host = `https://www.example.com`
+    const result = labelsFormatter(experimentAccession, host)
+
+    expect(result).toContain(`<img src="https://upload.wikimedia.org/wikipedia/commons/6/6a/External_link_font_awesome.svg"`)
```
ke4, line 82: Could we extract the full `img` tag to a `const` and use it in both tests, please?
That way when we change that image, we only have to change it in one place and I think the code would be cleaner.
Thanks

packages/scxa-marker-gene-heatmap/__test__/heatmapOptionsProvider.test.js
```diff
@@ -0,0 +1,96 @@
...
+    expect(result).toContain(`E-ANND-0001`)
+    expect(result).toContain(`href="https://www.example.com/experiments/E-ANND-0001"`)
+  })
+
+  it(`should return formatted label without AnnData icon for non-AnnData experiments`, () => {
+    const experimentAccession = `E-GEOD-1234`
```
ke4, line 88: Could we move the `host ` variable to outside of the tests as we are using it in both tests. 
Thanks

ke4 (changes requested):
Good work.
Only 2 tiny suggestions.
Many thanks

PR #176, opened 2024-09-26
173 overlapping names in the marker gene heat map

lingyun1010:
We had some UI issues in cell-type marker gene heatmap, while the cell type name is too long, and it will be overlapped with each other and users may not read it properly.

Given the discussion in the team meeting, we agreed that to make the cell type name at the right side column, tilted a bit like the title at the top row, and if it's still too long, we need to use `...` to shorten it.

Regarding the tests, actually, we just need to update the snapshots but we do not have any test configuration yet, this issue will be fixed in my another PR https://github.com/ebi-gene-expression-group/atlas-components/pull/178. I will suggest we do not worry the tests for this PR yet, and  we can upate the snapshots in the test refactoring PR instead.

### DIFF
--- a/packages/scxa-marker-gene-heatmap/src/MarkerGeneHeatmap.js
+++ b/packages/scxa-marker-gene-heatmap/src/MarkerGeneHeatmap.js
@@ -96,20 +96,19 @@ const MarkerGeneHeatmap = (props) => {
       const color = `#000000`
       const zIndex = 5
       const splitCellTypeLabel = splitPhrase(heatmapOptionsProvider[heatmapType].labelsFormatter(cellTypeOrClusterId))
+      const shortenCellTypeLabel = splitCellTypeLabel.length <= 3 ? splitCellTypeLabel.join(`<br/>`) : splitCellTypeLabel.slice(0, 3).join(`<br/>`) + `...`
       plotLines.push({
         color,
         width: 2,
         value: plotLineAxisPosition,
         zIndex,
         label: heatmapType !== `multiexperimentcelltypes` && {
-          // tilted and trim very long labels for long cell types
-          text: splitCellTypeLabel.length <= 3 ? splitCellTypeLabel.join(`<br/>`) : splitCellTypeLabel.slice(0, 3).join(`<br/>`) + `...`,
+          text: shortenCellTypeLabel,
           align: `right`,
           textAlign: `left`,
           rotation: heatmapType === `celltypes` ? -45 : 0,
           x: heatmapType === `celltypes` ? 3 : 0,
-          // No need to move very long labels slightly up as we tilted and trim it for long cell types
-          y: yOffset //- Math.max(splitCellTypeLabel.length - 3, 0) * heatmapRowHeight / 2
+          y: yOffset
         }
       })
     })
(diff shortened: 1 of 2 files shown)


packages/scxa-marker-gene-heatmap/src/MarkerGeneHeatmap.js
```diff
@@ -102,11 +102,14 @@ const MarkerGeneHeatmap = (props) => {
...
-          text: splitCellTypeLabel.join(`<br/>`),
+          // tilted and trim very long labels for long cell types
+          text: splitCellTypeLabel.length <= 3 ? splitCellTypeLabel.join(`<br/>`) : splitCellTypeLabel.slice(0, 3).join(`<br/>`) + `...`,
           align: `right`,
           textAlign: `left`,
-          x: 0,
-          y: yOffset - Math.max(splitCellTypeLabel.length - 3, 0) * heatmapRowHeight / 2 // Move very long labels slightly up
+          rotation: heatmapType === `celltypes` ? -45 : 0,
+          x: heatmapType === `celltypes` ? 3 : 0,
+          // No need to move very long labels slightly up as we tilted and trim it for long cell types
```
ke4, line 111: Could you remove this comment, please?
Having less or no comment in the code is always better as the comments unfortunately getting rotten as time goes.
If somebody change the code later, but leave the comment here, then it is going to confuse a new developer. It is better to leave less opportunity for this happening.
The best comment is always a non-existent comment. ;-)
We are using version control, so you can add this comment to your commit in the description session of the commit message. It is a much better place for this kind of comments.
How to do it: https://stackoverflow.com/questions/16122234/how-to-commit-a-change-with-both-message-and-description-from-the-command-li/36427485#36427485
Thanks

packages/scxa-marker-gene-heatmap/src/MarkerGeneHeatmap.js
```diff
@@ -102,11 +102,14 @@ const MarkerGeneHeatmap = (props) => {
...
+          y: yOffset //- Math.max(splitCellTypeLabel.length - 3, 0) * heatmapRowHeight / 2
```
ke4, line 112: I suspect that this comment was left here by accident. We are using version control, so we don't ned it.
Thanks

packages/scxa-marker-gene-heatmap/src/MarkerGeneHeatmap.js
```diff
@@ -102,11 +102,14 @@ const MarkerGeneHeatmap = (props) => {
         value: plotLineAxisPosition,
         zIndex,
         label: heatmapType !== `multiexperimentcelltypes` && {
-          text: splitCellTypeLabel.join(`<br/>`),
+          // tilted and trim very long labels for long cell types
+          text: splitCellTypeLabel.length <= 3 ? splitCellTypeLabel.join(`<br/>`) : splitCellTypeLabel.slice(0, 3).join(`<br/>`) + `...`,
```
ke4, line 106: Can you extract this code please to a `const` and add a well descriptive name to it?
That way you can remove the comment 1 line above.
Thanks

ke4 (changes requested):
Many thanks for your work.
I added some suggestions regarding comment usage and code cleaning.
Could you please consider them?
Thanks

upendrakumbham (approved):
LGTM!

ke4 (approved):
I have implemented my change requests ;-)

PR #178, opened 2024-10-01
177 chore refactor scxa marker geneheatmap enzyme tests with react 17

lingyun1010:
We used `enzyme` before to test marker-gene-heatmap component, which is based on React 16, and since we updated the react into 17, the tests had been suspended, in another word, we do not have any tests for this component currently.

To make a running jest tests again, I found out that `@testing-library/react` is very convinient and well-documented and keep updated to React. Hence, I refactor this component with TLR.

### DIFF
--- a/packages/scxa-marker-gene-heatmap/__test__/CalloutAlert.test.js
+++ b/packages/scxa-marker-gene-heatmap/__test__/CalloutAlert.test.js
@@ -1,30 +1,35 @@
 import React from 'react'
-import Enzyme from 'enzyme'
+import { render, screen } from '@testing-library/react'
+import '@testing-library/jest-dom' // for matcher like toBeInTheDocument
 import renderer from 'react-test-renderer'
-import { shallow } from 'enzyme'
-import Adapter from 'enzyme-adapter-react-16'
-
 import CalloutAlert from '../src/CalloutAlert'
 
-Enzyme.configure({ adapter: new Adapter() })
-
-describe(`CalloutAlert`, () => {
+describe('CalloutAlert', () => {
   const props = {
     error: {
-      description: `A human-readable description of the error, hopefully useful to the user`,
-      name: `Error name`,
-      message: `Error message`
+      description: 'A human-readable description of the error, hopefully useful to the user',
+      name: 'Error name',
+      message: 'Error message',
     }
   }
 
-  it(`prints all the relevant error information`, () => {
-    const wrapper = shallow(<CalloutAlert {...props} />)
-    expect(wrapper.text()).toMatch(props.error.description)
-    expect(wrapper.text()).toMatch(props.error.name)
-    expect(wrapper.text()).toMatch(props.error.message)
+  it('prints all the relevant error information', () => {
+    render(<CalloutAlert {...props} />)
+
+    expect(screen.getByText((content, element) => {
+      return content.includes(props.error.description)
+    })).toBeInTheDocument()
+
+    expect(screen.getByText((content, element) => {
+      return content.includes(props.error.name)
+    })).toBeInTheDocument()
+
+    expect(screen.getByText((content, element) => {
+      return content.includes(props.error.message)
+    })).toBeInTheDocument()
   })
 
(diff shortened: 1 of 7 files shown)


packages/scxa-marker-gene-heatmap/__test__/CalloutAlert.test.js
```diff
@@ -1,30 +1,35 @@
...
 import renderer from 'react-test-renderer'
-import { shallow } from 'enzyme'
-import Adapter from 'enzyme-adapter-react-16'
-
 import CalloutAlert from '../src/CalloutAlert'
 
-Enzyme.configure({ adapter: new Adapter() })
-
-describe(`CalloutAlert`, () => {
+describe('CalloutAlert', () => {
```
ke4, line 7: In other tests we are using backticks (`) and not single quotes (').
It would be better to use the same everywhere.
Thanks

packages/scxa-marker-gene-heatmap/__test__/CellTypeMarkerGeneHeatmap.test.js
```diff
@@ -1,92 +1,75 @@
...
-import Adapter from 'enzyme-adapter-react-16'
-
-import '@babel/polyfill'
+import { render, screen } from '@testing-library/react'
 import MarkerGeneHeatmap from '../src/MarkerGeneHeatmap'
 
-Enzyme.configure({ adapter: new Adapter() })
-
 describe(`MarkerGeneHeatmap`, () => {
   test(`creates plot lines for every cell type`, () => {
```
ke4, line 6: Shall we change `test` to `it` to unify the usage of a test definition at least inside a component?
Thanks

packages/scxa-marker-gene-heatmap/__test__/CellTypeMarkerGeneHeatmap.test.js
```diff
@@ -1,92 +1,75 @@
...
+        heatmapRowHeight={20}
+        hasDynamicHeight={false}
+        species="species"
+        heatmapType="celltypes"
+      />
+    )
 
-    expect(chartOptions.yAxis.plotLines).toHaveLength(3)
+    const plotLines = container.querySelectorAll('[class^="highcharts-plot-line "]')
+    expect(plotLines).toHaveLength(3)
```
ke4, line 46: `3` is a magic number. Why 3? I mean, I know why, but it would be better to use the size of the given input parameter.
Thanks

packages/scxa-marker-gene-heatmap/__test__/CellTypeMarkerGeneHeatmap.test.js
```diff
@@ -1,92 +1,75 @@
...
-      [
-        `downloadPNG`,
-        `downloadJPEG`,
-        `downloadPDF`,
-        `downloadSVG`,
-        `separator`,
-        `downloadCSV`,
-        `downloadXLS`
-      ]
+  test(`has data export options and a styled button`, () => {
```
ke4, line 49: You state 2 expectation in the definition of this test, but check only 1.
Also, you are stating: `styled button` but only check that a specific text is on the screen.
That is not telling us if that text is on a button and if that button has any style.
Could you fix this, please?
Thanks

ke4 (changes requested):
Good work!
I added some suggestions for change.
Could you please review them?
Thanks

packages/scxa-marker-gene-heatmap/__test__/CellTypeMarkerGeneHeatmap.test.js
lingyun1010: I modified all test to `test` insteadly. The test and it functions in Jest serve the same purpose but differ in their naming conventions and how they impact the readability of test descriptions. The choice between test and it is about readability rather than functionality. `test` is for more general tests, like UI or snapshots, `it` is more for unit functional tests.

ke4 (approved):
Thanks for your work.
LGTM!

PR #182, opened 2024-11-12
Refactor experiment page router component

lingyun1010:
This PR is the first part to create a front end component and could run the demo html page properly. The rest part would be create react routes tests thoroughly, which will be implemented in another PR, as we may need to work on this component for a urgent issue here https://github.com/orgs/ebi-gene-expression-group/projects/4/views/4?pane=issue&itemId=86174433&issue=ebi-gene-expression-group%7Catlas-web-single-cell%7C487.


packages/scxa-experiment-page/src/TSnePlotViewRoute.js
```diff
@@ -0,0 +1,338 @@
...
+          selectedPlotOptionLabel={search.plotOption ?
+                search.plotType ?
+                    getPlotOption(search.plotType.toLowerCase()) : getPlotOption(this.state.selectedPlotType.toLowerCase())
+                : this.state.selectedPlotOptionLabel
+          }
+          onChangePlotTypes={
+            (plotOption) => {
+              this.setState({
+                selectedPlotType: plotOption.value,
+                selectedPlotOption: Object.values(defaultPlotMethodAndParameterisation[plotOption.value])[0],
```
ke4, line 110: Could we please extract `defaultPlotMethodAndParameterisation[plotOption.value]` to a local variable?
Thanks

packages/scxa-experiment-page/src/TSnePlotViewRoute.js
```diff
@@ -0,0 +1,338 @@
...
+        main: () =>
+          <AnatomogramCellTypeHeatmapView
+            showIds={anatomogram[organWithMostOntologies]}
+            experimentAccession={experimentAccession}
+            species={species}
+            organ={organWithMostOntologies}
+            host={atlasUrl}
+          />
+      },
+      {
```
ke4, line 207: Is it possible to extract this routes element (currently 4) to their own files.
That might make this file a bit cleaner.

packages/scxa-experiment-page/src/TSnePlotViewRoute.js
```diff
@@ -0,0 +1,338 @@
...
+
+    const preferredK = this.props.selectedK ?
+        this.props.selectedK.toString() :
+          this.props.ks.length > 0 ?
+          this.props.ks[0].toString() : ``
+
+    const basename = URI(`experiments/${experimentAccession}${match.path}`, URI(atlasUrl).path()).toString()
+
+    const sideTabStyle = {overflow: `clip`, textOverflow: `ellipsis`}
+    return (
```
ke4, line 247: I would insert an empty line above this `return` statement to make the code more readable.
Thanks

packages/scxa-experiment-page/src/results/ontologyIds.json
```diff
@@ -0,0 +1,78 @@
+{
```
ke4, line 1: It is OK for now, but should not we store this in our PostgreSQL DB?
What do you think?

ke4 (changes requested):
Thanks for your hard work.
I added some clean code related comments, otherwise it looks good to me.

packages/scxa-experiment-page/src/TSnePlotViewRoute.js
```diff
@@ -0,0 +1,338 @@
...
+    const cellTypeValue = _first(_intersection(_map(this.props.metadata,`label`), this.props.initialCellTypeValues))
+    const search = URI(this.props.location.search).search(true)
+    this.state = {
+      selectedPlotType: Object.keys(this.props.defaultPlotMethodAndParameterisation)[0],
+      geneId: ``,
+      selectedPlotOption: Object.values(Object.values(this.props.defaultPlotMethodAndParameterisation)[0])[0],
+      selectedPlotOptionLabel: Object.keys(Object.values(this.props.defaultPlotMethodAndParameterisation)[0])[0] + ": " +
+          Object.values(Object.values(this.props.defaultPlotMethodAndParameterisation)[0])[0],
+      selectedColourBy: cellTypeValue ? cellTypeValue.toLowerCase() :
+          this.props.ks.length > 0 ? this.props.ks[Math.round((this.props.ks.length -1) / 2)].toString() : ``,
```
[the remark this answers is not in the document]
upendrakumbham: Same validation checks here as well before passing any function to fetch data
`Object.values(Object.values(this.props.defaultPlotMethodAndParameterisation)[0])[0]`

We are not sure about default drop down size is grater than 0

packages/scxa-experiment-page/src/TSnePlotViewRoute.js
upendrakumbham: Please wrap the code into a small function to improve readability.

packages/scxa-experiment-page/src/TSnePlotViewRoute.js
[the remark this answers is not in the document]
lingyun1010: > Same validation checks here as well before passing any function to fetch data `Object.values(Object.values(this.props.defaultPlotMethodAndParameterisation)[0])[0]`
> 
> We are not sure about default drop down size is grater than 0

Yes, but again, this is not the right place to do validation. We need the value here and I add `defaultPlotMethodAndParameterisation` as a required props. If we don't have this props, we should disable the tab, rather than keeping this value empty as `null`. In another words, the required props we do not validate in the component level anymore.

packages/scxa-experiment-page/src/TSnePlotViewRoute.js
```diff
@@ -0,0 +1,338 @@
...
+    const search = URI(location.search).search(true)
+
+    let plotTypeDropdown =
+      Object.keys(defaultPlotMethodAndParameterisation)
+        .map(plot => ({
+          plotType: plot,
+          plotOptions: plotTypesAndOptions[plot]
+        }))
+
+    let organWithMostOntologies = Object.keys(anatomogram)[0]
```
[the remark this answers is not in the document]
lingyun1010: This code snippet will not be resued, so I don't think it is good to create a function just because we need to clarify what is happening. I will add some comments instead to help understand.

packages/scxa-experiment-page/src/experiment-design/ExperimentDesignTable.js
```diff
@@ -0,0 +1,47 @@
...
+    <ReactTable
+      columns={
+        headers.map((headerGroup, ix)=> (
+          {
+            Header: headerGroup.name,
+            columns:
+              headerGroup.values.map((header, jx) => ({
+                aggregate: curry(aggregateText, 2)(header),
+                Header: header,
+                id: ix*1000 +jx +1,
```
ke4, line 31:
```suggestion
id: ix * 1000 + jx + 1,
```

packages/scxa-experiment-page/src/results/AnatomogramCellTypeHeatmapView.js
```diff
@@ -0,0 +1,124 @@
...
+  _addRemoveFromSelectIds(ids) {
+    if (ids.length === 0 || ids[0].startsWith(`link`)) {
+      this.setState({
+        selectIds: ids
+      })
+    } else {
+      this.setState({
+        selectIds: ids,
+        resource:
+            URI(`json/experiments/${this.props.experimentAccession}/marker-genes/cell-types`, this.props.host)
```
ke4, line 51: This could be extracted to a well named function and the same function could be used in the current line 67. 
Thanks

packages/scxa-experiment-page/src/results/AnatomogramCellTypeHeatmapView.js
```diff
@@ -0,0 +1,124 @@
...
+    })
+  }
+
+  render() {
+    const {host} = this.props
+
+    const heatmapArgs = {
+      host: host,
+      resource: this.state.resource,
+      species: `Homo sapiens`,
```
ke4, line 85: Is this always going to be `Homo sapiens` or it might change in the near future with the fly anatomograms? 
Thanks

packages/scxa-experiment-page/src/SupplementaryInformationRoute.js
```diff
@@ -0,0 +1,60 @@
...
+
+const SectionContent = ({type, props, atlasUrl}) => {
+  const Section = sectionTypeComponent[type]
+
+  return (Section && <Section {...{...props, atlasUrl}} />)
+}
+
+const SupplementaryInformationRoute = (props) => {
+  const annDataText =
+      'This methods summary is based on information provided by the authors or described in the associated ' +
```
ke4, line 19:
```suggestion
      `This methods summary is based on information provided by the authors or described in the associated ` +
      `publication, the data have not been re-analysed by Single Cell Expression Atlas`
```

ke4 (changes requested):
Thanks for your big refactor.
I added some tiny comments.
Could you please also check all of these files regarding for ESLint complains?
It was complaining in a few files. I did not want to add all of these into my review.
IntelliJ is going to show all of these.
Thanks a lot

packages/scxa-experiment-page/src/results/AnatomogramCellTypeHeatmapView.js
lingyun1010: yes, it will change to other species, if we would make it available.

packages/scxa-experiment-page/src/results/ontologyIds.json
lingyun1010: Yea or a gist should work too, so curator could modify it directly.

ke4 (approved):
LGTM
Thanks for your work!

upendrakumbham (approved):
LGTM!

PR #183, opened 2024-11-29
Improve marker gene right column label UI

lingyun1010:
As discribed in the related issue, we need to improve the right column labels for cell type marker gene heatmap.

### DIFF
--- a/packages/scxa-marker-gene-heatmap/src/MarkerGeneHeatmap.js
+++ b/packages/scxa-marker-gene-heatmap/src/MarkerGeneHeatmap.js
@@ -13,6 +13,8 @@ import _ from 'lodash'
 import URI from 'urijs'
 
 import heatmapOptionsProvider from './heatmapOptionsProvider'
+import formatTooltip from "./tooltip"
+import splitPhrase from "./splitPhrase"
 
 // initialise modules
 async function addModules() {
@@ -24,30 +26,6 @@ async function addModules() {
 }
 
 addModules()
-
-const splitPhrase = (phrase, maxLineLength = 12) => {
-  const splitPhrase = []
-
-  const words = phrase.split(/\s+/)
-
-  // Add first word to the first line if splitPhrase is empty
-  if (words.length > 0) {
-    splitPhrase.push(words.shift())
-  }
-
-  while (words.length > 0) {
-    const nextWord = words.shift()
-
-    // Check if adding the next word would exceed the maxLineLength
-    if (splitPhrase[splitPhrase.length - 1].length + nextWord.length + 1 <= maxLineLength) {
-      splitPhrase[splitPhrase.length - 1] = `${splitPhrase[splitPhrase.length - 1]} ${nextWord}`
-    } else {
-      splitPhrase.push(nextWord)
-    }
-  }
-
-  return splitPhrase
-}
 
 Highcharts.SVGRenderer.prototype.symbols.download = (x, y, w, h) => [
   // Arrow stem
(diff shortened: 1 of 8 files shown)


packages/scxa-marker-gene-heatmap/src/MarkerGeneHeatmap.js
```diff
@@ -26,12 +26,20 @@ async function addModules () {
 addModules()
 
 const splitPhrase = (phrase, maxLineLength = 12) => {
-  const splitPhrase = [``]
+  const splitPhrase = []
 
   const words = phrase.split(/\s+/)
+
+  // Add first word to the first line if splitPhrase is empty
```
ke4, line 33: At this point `splitPhrase` is always empty. You defined it as empty in line 29 and nothing changes it.
The comment does not reflect what is happening in the code.
It adds the first word to the first line if `words` has any content and not when `splitPhrase` is empty.
Could you please remove this comment?
Thanks

packages/scxa-marker-gene-heatmap/src/MarkerGeneHeatmap.js
```diff
@@ -26,12 +26,20 @@ async function addModules () {
...
+  if (words.length > 0) {
+    splitPhrase.push(words.shift())
+  }
+
   while (words.length > 0) {
     const nextWord = words.shift()
-    if (splitPhrase[splitPhrase.length - 1].length + nextWord.length < maxLineLength) {
+
+    // Check if adding the next word would exceed the maxLineLength
```
ke4, line 41: Could you please remove the comment and extract the if condition to a well named method?
That way we don't need the comment and the code is more readable.
Thanks

packages/scxa-marker-gene-heatmap/src/MarkerGeneHeatmap.js
```diff
@@ -96,19 +104,29 @@ const MarkerGeneHeatmap = (props) => {
...
+          text: splitCellTypeLabel,
           align: `right`,
           textAlign: `left`,
           rotation: heatmapType === `celltypes` ? -45 : 0,
           x: heatmapType === `celltypes` ? 3 : 0,
-          y: yOffset
+          y: yOffset,
+          formatter: () => {
+            const shortenCellTypeLabel = numberOfRows === 1 ?
+              splitCellTypeLabel.slice(0, 1).join(`<br/>`) + `...` :
```
ke4, line 121: Is this code always put `...` at the end of a label when `numberOfRows === 1` even if the whole label text fits into the space?

packages/scxa-marker-gene-heatmap/src/MarkerGeneHeatmap.js
```diff
@@ -121,7 +139,66 @@ const MarkerGeneHeatmap = (props) => {
...
+
+          // Find the label by className
+          chart.yAxis[0].plotLinesAndBands.forEach(plotLine => {
+            if (plotLine.label) {
+              // Get the rendered SVG label element
+              const label = plotLine.label.element
+
+              label.addEventListener(`mouseover`, function () {
+                const tooltip = document.createElement(`div`)
+                tooltip.className = `custom-tooltip`
```
ke4, line 155: Could we please extract the styling of the tooltip to a method?
Thanks

packages/scxa-marker-gene-heatmap/src/MarkerGeneHeatmap.js
```diff
@@ -121,7 +139,66 @@ const MarkerGeneHeatmap = (props) => {
...
+                tooltip.style.zIndex = 1000
+
+                // Add styling for better display
+                tooltip.style.maxWidth = `60px` // Restrict width
+                tooltip.style.wordWrap = `break-word` // Enable word wrapping
+                tooltip.style.boxShadow = `0 2px 5px rgba(0,0,0,0.3)` // Add subtle shadow for visibility
+
+                tooltip.innerText = plotLine.options.label.text
+
+                // Position tooltip
```
ke4, line 170: Could we please extract the positioning of the tooltip to a method?
Thanks

ke4 (changes requested):
Thanks for your work!
I have added mainly code cleaning related comments.
I also have a concern that `...` would always added to the label when we have only 1 row to display it.
Thanks

packages/scxa-marker-gene-heatmap/src/MarkerGeneHeatmap.js
lingyun1010: I added another condition if the splitCellTypeLabel just has one line then we don't add `...`.

packages/scxa-marker-gene-heatmap/src/MarkerGeneHeatmap.js
```diff
@@ -118,7 +96,9 @@ const MarkerGeneHeatmap = (props) => {
           y: yOffset,
           formatter: () => {
             const shortenCellTypeLabel = numberOfRows === 1 ?
-              splitCellTypeLabel.slice(0, 1).join(`<br/>`) + `...` :
+              splitCellTypeLabel.length === 1 ?
```
ke4, line 99: Could we refactor the 2 different method when we populate the plotLines variable, please (line 65 & 85)?
I would extract those lines to 2 well named methods.
Thanks

ke4 (changes requested):
Thanks for your work.
Could you please review my suggestion for code cleaning?
Thanks

ke4 (approved):

PR #185, opened 2024-12-09
Fix the initialization issue on selected colour by category

lingyun1010:
Gene expression heatmap has a wrong initialization heatmap type. https://www-test.ebi.ac.uk/gxa/sc/experiments/E-MTAB-10448/results/marker-genes

This will happend to experiments do not have metadatas. Check the related issue for more information.


ke4 (changes requested):
Good catch!
We need to update / bump this package's version in the `package.json` file.
Thanks

ke4 (approved):
LGTM

PR #199, opened 2025-03-04
Fix security vulnerabilities in gene search form

lingyun1010:
There are three tests regarding autocomplete still failing, and we decided to leave them for now to finish the security vulnerability task and upgration to webpack 5.


ke4 (changes requested):
We agreed in other components that we are not using localhost for the demo, but replace them with staging URL.
Could you please fix it?
Thanks

lingyun1010:
> We agreed in other components that we are not using localhost for the demo, but replace them with staging URL. Could you please fix it? Thanks

We do use `localhost` in demo.

amnonkhen (approved):

PR #201, opened 2025-03-13
read lists not objects

amnonkhen:
read the `/experiment`'s series data as a list of lists rather than a list of objects as part of performance improvements.

see ebi-gene-expression-group/atlas-web-single-cell#503


lingyun1010 (dismissed):
LGTM!

PR #205, opened 2025-03-21
Fix security vulnerabilities in scxa experiment page

lingyun1010:


packages/scxa-experiment-page/package.json
```diff
@@ -53,12 +54,16 @@
...
     "clean-webpack-plugin": "^3.0.0",
     "eslint": "^8.17.0",
+    "eslint-plugin-cypress": "^2.15.2",
     "eslint-plugin-jest": "^28.8.3",
     "file-loader": "^6.1.0",
-    "image-webpack-loader": "^7.0.0",
-    "webpack": "^4.44.1",
-    "webpack-cli": "^3.3.12",
-    "webpack-dev-server": "^3.11.0",
+    "image-webpack-loader": "^4.5.0",
```
ke4, line 60: Why are we downgrading this dependency?

ke4 (changes requested):
This package is lack of tests.  Could you tell me what is the reason?
I saw that you downgraded `image-webpack-loader` dependency.
Is that necessary?
Thanks

packages/scxa-experiment-page/package.json
lingyun1010: I think that's an automatic fix from npm audit fix. I will bring it back to version 7.

lingyun1010:
A new ticket is added regarding add tests https://github.com/orgs/ebi-gene-expression-group/projects/4/views/18?pane=issue&itemId=103277233&issue=ebi-gene-expression-group%7Catlas-components%7C208

PR #207, opened 2025-03-21
Fix security vulnerability in atlas homepage card

lingyun1010:


ke4 (changes requested):
Test execution fails at the end and the branch has merge conflicts.
Could you please take a look and fix them?
Thanks

lingyun1010:
The coverall error is not for jest test. It's due to we do not have the original repository since we moved all components to lerna monorepo. I removed the `posttest` script for now. If we still want `coverall`, we should add more tickets regarding this.

amnonkhen (approved):

